PR #4, opened 2021-12-19
start using apollo watch() to drive UI + add KMP-NativeCoroutines for iOS

joreilly:
@martinbonnin seems like using `watch()` to drive UI is probably preferred approach? (as per https://www.apollographql.com/docs/kotlin/caching/query-watchers/) ....though I might need to think a bit more about lifecycle related implications for how I have code setup for this right now.


martinbonnin:
Yep, `watch()` is nice because it makes sure the UI is always using the latest version of the cache and all screens use the same source of truth. Note that it's watching the cache, not the backend so if the data changes on the backend, something still has to trigger a refetch manually. For backend updates, we'd need subscriptions.

PR #6, opened 2021-12-19
initial session details screen on iOS

joreilly:


shared/src/commonMain/kotlin/dev/johnoreilly/kikiconf/model/Session.kt
```diff
@@ -3,7 +3,7 @@ package dev.johnoreilly.kikiconf.model
 import dev.johnoreilly.kikiconf.GetSessionQuery
 import dev.johnoreilly.kikiconf.GetSessionsQuery
 
-data class Session(val id: String, val title: String, val description: String, val tags: List<String>)
+data class Session(val id: String, val title: String, val desc: String, val tags: List<String>)
```
joreilly, line 6: looks. like `description` is a keyword in swift

PR #7, opened 2021-12-19
A few minor miscelaneous changes

martinbonnin:


androidApp/build.gradle.kts
```diff
@@ -28,20 +28,16 @@ android {
             isMinifyEnabled = false
         }
     }
-
-    compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
-    }
```
martinbonnin, line 35: This is by default in recent AGP

androidApp/build.gradle.kts
```diff
@@ -28,20 +28,16 @@ android {
...
-
-    kotlinOptions {
-        jvmTarget = "1.8"
-    }
```
martinbonnin, line 39: This is by default in recent Kotlin Plugin versions

buildSrc/build.gradle.kts
```diff
@@ -1,11 +1,7 @@
 repositories {
-    jcenter()
+    mavenCentral()
 }
 
 plugins {
-    `kotlin-dsl`
+    `embedded-kotlin`
```
martinbonnin, line 6: I like `embedded-kotlin` better because it doesn't apply the SamWithReceiver plugin https://kotlinlang.org/docs/sam-with-receiver-plugin.html, which can be surprising. Both work though so feel free to revert

PR #22, opened 2022-05-02
start using startInstant/endInstant

joreilly:


martinbonnin (approved):

shared/src/commonMain/graphql/schema.json
```diff
@@ -338,6 +344,18 @@
...
+            "isDeprecated": false,
+            "type": {
+              "kind": "NON_NULL",
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Instant"
+              }
+            },
+            "args": []
+          },
```
martinbonnin, line 358: Nit: that could use SDL for better readability

joreilly: ah, keep forgetting about that.....this is what I'm using right now.....can't remember, is this different option here?

```
./gradlew downloadApolloSchema --endpoint=https://kiki-conf.ew.r.appspot.com/graphql --schema="shared/src/commonMain/graphql/schema.json"
```

joreilly: I'll  merge this for now and we can do subsequent update for this.

martinbonnin: If you name your schema `schema.graphqls`, it will be converted to SDL automagically:

```
./gradlew downloadApolloSchema --endpoint=https://kiki-conf.ew.r.appspot.com/graphql 
--schema="shared/src/commonMain/graphql/schema.graphqls"
```

PR #24, opened 2022-05-02
use config time zone info from server

joreilly:
Addresses https://github.com/joreilly/Confetti/issues/23


joreilly:
As per slack discussion there's definitely more work needed here but probably ok to merge this as is for now..

PR #42, opened 2022-06-07
Server: make the session list paginated

BoD:


server/src/main/kotlin/fr/androidmakers/server/CachedData.kt
```diff
@@ -116,3 +135,11 @@ object CachedData {
...
     }
 }
+
+fun String.encodeBase64(): String {
+    return Base64.getEncoder().encodeToString(toByteArray())
+}
+
+fun String.decodeBase64(): String {
+    return String(Base64.getDecoder().decode(this))
+}
```
martinbonnin, line 145: Why encode/decode to base64?

martinbonnin (approved):
One question but looks good otherwise 👍

server/src/main/kotlin/fr/androidmakers/server/CachedData.kt
BoD: It's not really needed but I think it's a good practice for them to be "opaque" - it conveys the fact that they're really cursors and not ids 😊.  Some [reference](https://graphql.org/learn/pagination/#:~:text=In%20general%2C%20we%27ve,base64%20encoding%20them.).

PR #53, opened 2022-06-30
DO NOT MERGE Pagination for the session list using the new Apollo Kotlin experimental APIs

BoD:
Do not merge, this PR is only there to have a look at what it would be to use the new experimental APIs in Apollo Kotlin to handle pagination in the normalized cache.

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionView.kt
@@ -1,10 +1,9 @@
 package dev.johnoreilly.confetti.sessions
 
-import android.annotation.SuppressLint
 import androidx.compose.foundation.*
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.lazy.LazyColumn
-import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.*
 import androidx.compose.material.icons.Icons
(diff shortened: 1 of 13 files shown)


androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionView.kt
```diff
@@ -25,32 +24,45 @@ import dev.johnoreilly.confetti.fragment.SessionDetails
...
     val sessions by viewModel.sessions.collectAsState(emptyList())
 
     val enabledLanguages by viewModel.enabledLanguages.collectAsState(emptySet())
 
     Scaffold(
-        topBar = { TopAppBar (
-            title = { Text("Sessions") },
-            // TODO need to figure out how we want to generally handle languages
+        topBar = {
+            TopAppBar(
```
BoD, line 38: Sorry about reformatting! I compulsively press alt-cmd-L all the time...

server/src/main/kotlin/fr/androidmakers/server/model/model.kt
```diff
@@ -32,8 +32,8 @@ data class SessionEdge(
 data class PageInfo(
     val hasPreviousPage: Boolean,
     val hasNextPage: Boolean,
-    val startCursor: String,
-    val endCursor: String,
+    val startCursor: String?,
```
BoD, line 35: Turns out, lists can be empty 😅

shared/src/commonMain/graphql/extra.graphqls
```diff
@@ -0,0 +1,6 @@
+extend type Query
+@typePolicy(embeddedFields: "sessions")
+@fieldPolicy(forField: "sessions", paginationArgs: "first, after")
```
BoD, line 3: This is the current way to make pagination work - ideally we'll find a way to make this less involved/verbose.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/di/Koin.kt
```diff
@@ -29,14 +29,103 @@ fun commonModule() = module {
...
+            apolloResolver = FieldPolicyApolloResolver,
+            recordMerger = FieldRecordMerger(CursorPaginationFieldMerger()),
+            writeToCacheAsynchronously = true
+        )
         .build()
 }
+
+@OptIn(ApolloExperimental::class)
+@Suppress("UNCHECKED_CAST")
+private class CursorPaginationMetadataGenerator(private val connectionTypes: Set<String>) :
```
BoD, line 54: This (and `CursorPaginationFieldMerger`) would probably reside in the library, as it should work the same in all Relay-based pagination APIs. For now I copied/pasted it from [the test in the library](https://github.com/apollographql/apollo-kotlin/pull/4237/files?w=1#diff-6b5012b1b3780a2db4c73cfef61a99c727e36cab2197c5e5f426bf01fa0e0545R314).

BoD:
Closing for now! See also #57 for an example of using this with Jetpack Paging.

PR #55, opened 2022-07-05
Add a favorite session list, and ability to favorite a session

BoD:
Another exploratory PR (not meant to be merged!)

This explores the case where a mutation updates an item that is part of a paginated list:
- a filter on the Session list to filter on favorites only
- a button on the Session details allows to add to / remove from favorites

Note: Favorites are mocked on the server (in-memory only and shared for all users)


joreilly:
@BoD nice changes...cool to see app include a mutation!  you might have seen already but look like PR build is failing.

BoD:
I think it was because the snapshots were not published yet 😅  Restarted the job.

joreilly:
ah, went to try this out after merging and looks like it was merged in to `session-list-pagination` (and not `main`)....not sure if that was intention?

joreilly:
I wonder how much would be involved in doing something like approach
firebase anonymous authentication uses for now

PR #57, opened 2022-07-20
Use Jetpack Paging

BoD:
Demonstrates how to use the Paging component combined with Apollo Kotlin's recent pagination facilities.

Disclaimer: it seems to work well but I must admit I'm not 100% confident everything is correct, as it took me a good while of trial/error to settle on this.

The documentation is [here](https://developer.android.com/topic/libraries/architecture/paging/v3-network-db) but to summarize:
- a `SessionRemoteMediator` is responsible for downloading pages from the network, and to store them in the cache. In our case the storing is actually done automatically by the Apollo client.
- a `SessionPagingSource` is responsible for querying the cache for specific pages
- it is also responsible for invalidating _itself_ when it knows the data it has returned is stale. In our case we do this by using `watch` on the sessions.
- the UI observes the data returned by the PagingSource, via a `Pager` held by the ViewModel

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -46,7 +46,7 @@ dependencies {
     implementation("androidx.activity:activity-compose:1.4.0")
     implementation("com.google.android.material:material:1.4.0")
 
-    with (Compose) {
+    with(Compose) {
         implementation(compiler)
         implementation(ui)
         implementation(uiGraphics)
@@ -59,9 +59,14 @@ dependencies {
         implementation(coilCompose)
     }
 
-    with (Koin) {
+    with(Koin) {
         implementation(core)
         implementation(android)
+        implementation(compose)
+    }
+
+    with(Paging) {
+        implementation(runtime)
         implementation(compose)
     }
 }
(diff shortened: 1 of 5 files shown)


BoD:
Woops, sorry about iOS, I should have known.

About no data showing up, is your server running locally (`./gradlew :server:bootRun`)? On this branch I use the local server because of changes that are not published to the 'prod' one.

joreilly:
>Good reminder that I must update github workflow so that it does some kind of build verification of iOS client

and looks like @martinbonnin has done just that 👍   https://github.com/joreilly/Confetti/pull/60/files

martinbonnin:
Working on it... Too afraid of breaking things when touching the code so at least having a CI will give a better level of confidence 🤞

martinbonnin:
@BoD want to get this through the finish line or should we close it?

BoD:
Closing for now, it was a way to test/demo the cache pagination incubating features, but I don't think it's actually useful for the app (plus using something other than Jetpack Paging may make more sense for a KMP project) :).

PR #59, opened 2022-07-25
Server CI

martinbonnin:
Deploy the server to app engine on pushes to main (well and pull_requests for testing but I'll remove this quickly)


joreilly:
nice!

PR #96, opened 2022-09-30
Frenchkit 2022

martinbonnin:
A bit late to the party but it's there 🎊 

[androidApp-debug.apk](https://storage.googleapis.com/martin-public/androidApp-debug.apk)

Screenshot_20220930_154242

### DIFF
--- a/backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/FrenchKit.kt
+++ b/backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/FrenchKit.kt
@@ -1,0 +1,99 @@
+package dev.johnoreilly.confetti.backend.import
+
+import dev.johnoreilly.confetti.backend.datastore.*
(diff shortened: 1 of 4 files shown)


backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/FrenchKit.kt
```diff
@@ -0,0 +1,99 @@
...
+            rooms = rooms,
+            speakers = speakers,
+            partnerGroups = emptyList(),
+            config = DConfig(
+                timeZone = "Europe/Paris"
+            ),
+            venues = listOf(
+                DVenue(
+                    id = "main",
+                    name = "Cité des Congrès de Nantes",
```
BoD, line 86: I think it's not there though? :)

martinbonnin: Not quite there :)

PR #113, opened 2022-10-14
iOS pull to refresh

joreilly:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -92,21 +94,22 @@ class ConfettiRepository : KoinComponent {
         appSettings.updateEnableLanguageSetting(language, checked)
     }
 
-    fun refresh(fetchPolicy: FetchPolicy = FetchPolicy.NetworkOnly) {
+    suspend fun refresh(networkOnly: Boolean = true) {
```
martinbonnin, line 97: I initially made this non-suspend as I was thinking of it of "trigger" and the end of the operation would be that `isRefreshing` emits false. If we're making this suspend, can we get rid of `isRefreshing`?

joreilly: hmm, probably....I'll try that out here.  btw main reason I made it suspend so that it would be "part of" the same async task on iOS in terms of way refreshable etc works

martinbonnin: Removed `isRefreshing` there: https://github.com/joreilly/Confetti/pull/116

PR #120, opened 2022-10-15
Properly set the name in the DB

martinbonnin:


backend/datastore/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/datastore/DataStore.kt
```diff
@@ -313,13 +313,14 @@ class DataStore {
                 .setKind(KIND_CONFIG)
                 .newKey(THE_CONFIG)
         )
+            .set("name", name.toValue())
```
martinbonnin, line 316: That's the proper fix

backend/datastore/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/datastore/DataStore.kt
```diff
@@ -313,13 +313,14 @@ class DataStore {
...
             .set("timeZone", timeZone.toValue())
             .build()
     }
 
     private fun Entity.toConfig(): DConfig {
         return DConfig(
-            name = getString("name"),
+            name = getString("name") ?: "",
```
martinbonnin, line 323: That's a fail safe to avoid future crashes

PR #122, opened 2022-10-15
Upload the APK to Github Actions

martinbonnin:
IPA still TODO


androidApp/build.gradle.kts
```diff
@@ -14,6 +14,14 @@ android {
         versionName = "1.0"
     }
 
+    signingConfigs {
+        create("confetti") {
+            keyAlias = "confetti"
+            keyPassword = "confetti"
+            storeFile = file("confetti.keystore")
+            storePassword = "confetti"
```
martinbonnin, line 22: Putting passwords in clear means anyone can impersonate the app but I think this is fine at this point :)

androidApp/build.gradle.kts
```diff
@@ -26,6 +34,10 @@ android {
     buildTypes {
         getByName("release") {
             isMinifyEnabled = false
+            signingConfig = signingConfigs.getByName("confetti")
+        }
+        getByName("debug") {
+            signingConfig = signingConfigs.getByName("confetti")
```
martinbonnin, line 40: Changing the keystore here means you'll have to remove your current debug app before installing a new one

PR #140, opened 2022-10-28
offline support updates

joreilly:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -53,47 +52,25 @@ class ConfettiRepository : KoinComponent {
         Conference("droidconlondon2022", "Droidcon London 2022"),
     )
 
-    private sealed interface EverythingResult
```
joreilly, line 56: With other changes being made (e.g. no longer, ,at least for now!, setting that error result) and loading being handled in VM layer it seemed like we weren't getting as much value from this as before.....but can certainly add something like it back again if needed.

martinbonnin: Sounds good 👍 

Philosophical question: where do you draw the line between the VM responsibilities and the Repository responsibilities :)?

joreilly: Probably some grey areas there (and probably some stuff still in app right now that could get moved in either direction potentially) but I guess VM should primarily be responsible for creating UI level observable state and as conduit for actions from UI....including I think Loading state for example.  Probably some updates needed still around error handling....if you look at https://developer.android.com/topic/architecture/data-layer#expose-errors  for example I think guidance is that repository should throw exceptions when failures occur.

PR #159, opened 2022-12-16
updates for androiid play store

joreilly:


martinbonnin:
Is androiid a bit like Androïde?

PR #177, opened 2022-12-22
Use OkHttp 5 - coroutines and happy eyeballs

yschimke:
For discussion - happy to close if not in scope.  But you seem to use modern versions a lot.

Uses a single instance of OkHttp across Apollo and Coil.  Uses clean coroutines code in all places also, and happy eyeballs to speed up IPv6/IPv4 related issues.

### DIFF
--- a/backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
+++ b/backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
@@ -2,46 +2,55 @@ package dev.johnoreilly.confetti.backend.import
 
 import com.charleskorn.kaml.*
 import dev.johnoreilly.confetti.backend.datastore.*
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import kotlinx.datetime.*
 import kotlinx.serialization.json.Json
 import net.mbonnin.bare.graphql.asList
 import net.mbonnin.bare.graphql.asMap
 import net.mbonnin.bare.graphql.asString
 import net.mbonnin.bare.graphql.toAny
+import okhttp3.executeAsync
+import okhttp3.HttpUrl.Companion.toHttpUrl
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.minutes
 
 private val timeZone = "Europe/Paris"
-private val okHttpClient = OkHttpClient()
+
+private val okHttpClient = OkHttpClient.Builder()
+    .fastFallback(true)
+    .build()
+
 private val baseUrl = "https://raw.githubusercontent.com/GDG-Nantes/Devfest2022/master/"
 private val json = Json {
     ignoreUnknownKeys = true
 }
 
 object DevFestNantes {
-    private fun getUrl(url: String): String {
-        return Request.Builder()
-            .url(url)
-            .build()
-            .let {
-                okHttpClient.newCall(it).execute().also {
-                    check(it.isSuccessful) {
-                        "Cannot get $url: ${it.body?.string()}"
-                    }
-                }
-            }.body!!.string()
+    private suspend fun getUrl(url: String): String {
+        val request = Request(url.toHttpUrl())
+        val response = okHttpClient.newCall(request).executeAsync()
+
+        check(response.isSuccessful) {
+            "Cannot get $url: ${response.body.string()}"
+        }
+
+        return withContext(Dispatchers.IO) {
+            response.body.string()
+        }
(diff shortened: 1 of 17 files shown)


joreilly:
will defer to @martinbonnin on this

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
```diff
@@ -2,46 +2,56 @@ package dev.johnoreilly.confetti.backend.import
...
+    private suspend fun getUrl(url: String): String {
+        val request = Request(url.toHttpUrl())
+        val response = okHttpClient.newCall(request).executeAsync()
+
+        return response.use {
+            check(response.isSuccessful) {
+                "Cannot get $url: ${response.body.string()}"
+            }
+
+            withContext(Dispatchers.IO) {
```
martinbonnin, line 40: What's the reason we're using `Dispatchers.IO` for reading the body but not the initial response?

yschimke: executeAsync is suspending and so safe for any dispatcher. But the best pattern for IO we have is IO dispatcher and sync read calls

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
```diff
@@ -20,7 +20,6 @@ import kotlin.time.Duration.Companion.minutes
 private val timeZone = "Europe/Paris"
 
 private val okHttpClient = OkHttpClient.Builder()
-    .fastFallback(true)
```
martinbonnin: Why not keeping them?

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
martinbonnin: Can we enforce the same dispatcher for both? Looks like in the current state, `executeAsync` uses the OkHttp threadpool and `body.string()` will use the `Dispatchers.IO`? Feels weird to hop thread 3 times for this. Unless somehow OkHttp threadpool == Dispatchers.IO ?

yschimke: I'm not sure we can.  `executeAsync` is this:

```
actual suspend fun Call.executeAsync(): Response = suspendCancellableCoroutine { continuation ->
  continuation.invokeOnCancellation {
    this.cancel()
  }
  this.enqueue(object : Callback {
    override fun onFailure(call: Call, e: IOException) {
      continuation.resumeWithException(e)
    }

    override fun onResponse(call: Call, response: Response) {
      continuation.resume(value = response, onCancellation = { call.cancel() })
    }
  })
}
```

So whatever Dispatcher you put it on, you end up in a callback with processing being done by OkHttp threads.  I don't think moving this onto Dispatchers.IO actually improves anything. 

But ultimately, I'm happy to put it all under Dispatchers.IO, because I doubt it's harmful, I just didn't reason it being an improvement.

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
yschimke: It's defaulted on in 5.0.0-alpha.11

https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-500-alpha11

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
martinbonnin: >  I don't think moving this onto Dispatchers.IO actually improves anything.

Agreed 👍 
Overall, this is all mostly theoretical for a CLI script that isn't doing a lot more. I was mainly curious if there's a reason to introduce Dispatchers.IO when everything could be run from OkHttp threads. But I think everything works at this point, especially because nothing is run in parallel.

backend/service-import/src/jvmMain/kotlin/dev/johnoreilly/confetti/backend/import/DevFestNantes.kt
martinbonnin: Nice!

martinbonnin (approved):
Very cool 👍

Issue #194, opened 2023-01-13
Change Wear navigation.


yschimke:
Settings placement is problematic.  https://github.com/joreilly/Confetti/pull/193

Requires a lot of scrolling, and is at the bottom of each date in the pager.

Guidelines state apps should be linear https://developer.android.com/training/wearables/views/navigation#design.

A better design would be 

Home page - Todays sessions filtered (now & next?) with links to each date in full plus settings
Date page - no longer part of the pager.

PR #195, opened 2023-01-13
Refactor Wear Nav

yschimke:
https://github.com/joreilly/Confetti/issues/194

Still a bit of a draft.  Should centralise logic to show the relevant (now and next?, favourites?) sessions on the home screen.

Also how to usefully show succinct dates and times.

[nav.webm](https://user-images.githubusercontent.com/231923/212308244-08ab604d-54e2-4bb1-a24a-74e57a37884f.webm)


joreilly:
Dates are always fun!  They're also somewhat contextual....looking in to future it might be good to see actual dates but if conference is currently on then I like having day names like you have here.

joreilly:
Maybe obvious point but I think main benefit for app like this is probably people actually attending conference so having days by default might be good.

joreilly:
@yschimke sounded like you wanted to make some more changes before we merge this.....or will I merge it now?

PR #198, opened 2023-01-14
Fix the Wear Session Details view.

yschimke:
Strip bad characters, split on paragraphs, and fix layout in the ScalingLazyColumn.

[text.webm](https://user-images.githubusercontent.com/231923/212468655-d2271576-14f5-46ba-90e4-828cb33d2907.webm)


wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessiondetails/SessionDetailsView.kt
```diff
@@ -30,59 +32,91 @@ import java.time.ZoneOffset
...
-                    }
-                )
+                SessionSpeakerInfo(speaker = speaker.speakerDetails)
             }
         }
     }
 }
 
-@RequiresApi(Build.VERSION_CODES.O)
+private fun SessionDetails?.descriptionParagraphs(): List<String> {
```
yschimke, line 85: @joreilly any thoughts on how to handle this? During the load?

joreilly: @martinbonnin I think you were hoping to "santize" this data at some point? (this is specific I think to fosdem info?)

martinbonnin: Indeed, this is only for Fosdem data that is HTML. I can move that logic to the backend

yschimke: https://github.com/joreilly/Confetti/issues/199

PR #221, opened 2023-01-29
Update for Wear Compose 1.2 alphas

yschimke:
Includes package migration described here https://developer.android.com/jetpack/androidx/releases/wear-compose#version_12_2


wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessions/navigation/sessionsDestination.kt
```diff
@@ -38,6 +38,10 @@ fun NavGraphBuilder.sessionsGraph(
     ) {
         val date = SessionsDestination.fromNavArgs(it.backStackEntry)
 
-        SessionsRoute(date = date, navigateToSession = navigateToSession)
+        SessionsRoute(
+            date = date,
+            navigateToSession = navigateToSession,
+            columnState = it.columnState
```
yschimke, line 44: This was causing the time text to not scroll out of the way

PR #222, opened 2023-01-29
Add optional crashlytics

### LINKED ISSUE #220: Crashlytics reporting
For me, part of the interest in the app is a real app, with legitimate real users.

yschimke:
Register a project at console.firebase.google.com and copy google-services.json into wearApp and androidApp.

fixes #220


yschimke:
Oops, sorry. I should set it up so it doesn't need to be there. 

But committing also works?

martinbonnin:
I like committing, this way everyone shares the same configuration and it's less things to debug. The keys are public in the apk anyways.

PR #235, opened 2023-02-04
Automate Wear Screenshots

yschimke:
First step towards keeping screenshots automatically inline with the app design.

```
$ fastlane wearScreenshots
```

Next steps would be mobile, then eventually uploading to the play store on each release.


joreilly:
Nice!

PR #237, opened 2023-02-06
add refresh option to wear app

joreilly:
Can maybe find better place for this but just to have option for now in case data is stale.


yschimke:
LGTM - thanks!

PR #245, opened 2023-02-14
[RFC] Refactor to make conference un-special

yschimke:
Early draft for discussion.

Changes
- Use non blocking settings API
- Use DataStore<Preferences> as the permanent settings
- Make conference a part of the navigation hierarchy

Context: I was looking at how to sync the Wear conference directly from the Mobile setting.  As part of that, I wanted to have it able to change immediately, and prompt to navigate the user across conferences.  This syncing can be put behind the DataStore API, so I wanted to use that API consistently.  

Before I finish off this, I wanted to check on early feedback, including whether to abandon and rethink the idea. Or split up.

But overall I think it's a nicer design for the app.  Allows linking directly into another conference, say from a push notification. And removes a lot of blocking logic during startup.

### DIFF
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
@@ -1,0 +1,42 @@
+package dev.johnoreilly.confetti
+
+import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.cache.normalized.api.MemoryCacheFactory
+import com.apollographql.apollo3.cache.normalized.normalizedCache
+import com.apollographql.apollo3.cache.normalized.sql.SqlNormalizedCacheFactory
+import dev.johnoreilly.confetti.di.getDatabaseName
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+
+class ApolloClientCache : KoinComponent {
+    val _clients = mutableMapOf<String, ApolloClient>()
+    val mutex = Mutex(false)
+
+    suspend fun getClient(conference: String): ApolloClient {
+        return mutex.withLock {
+            _clients.getOrPut(conference) {
+                clientFor(conference, conference != "all")
+            }
+        }
+    }
+
+    private fun clientFor(
+        conference: String,
+        writeToCacheAsynchronously: Boolean
+    ): ApolloClient {
+        val sqlNormalizedCacheFactory = SqlNormalizedCacheFactory(getDatabaseName(conference))
+        val memoryFirstThenSqlCacheFactory = MemoryCacheFactory(10 * 1024 * 1024)
+            .chain(sqlNormalizedCacheFactory)
+
+        return get<ApolloClient.Builder>()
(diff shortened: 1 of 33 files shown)


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.apollographql.apollo3.cache.normalized.api.MemoryCacheFactory
+import com.apollographql.apollo3.cache.normalized.normalizedCache
+import com.apollographql.apollo3.cache.normalized.sql.SqlNormalizedCacheFactory
+import dev.johnoreilly.confetti.di.getDatabaseName
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+
+class ApolloClientCache : KoinComponent {
```
joreilly, line 13: does this need to be a `KoinComponent` ?

yschimke: There is a call below to 

```
return get<ApolloClient.Builder>()
```

yschimke: Suggestions?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -1,141 +1,103 @@
...
     }
 
-    val sessionsMap: Flow<Map<LocalDate, List<SessionDetails>>> = sessions.map {
-        it.groupBy {
-            it.startInstant.toLocalDateTime(
-                TimeZone.of(
-                    conferenceData.value?.config?.timezone ?: ""
-                )
-            ).date
+    fun conferenceDataFlow(): Flow<GetConferenceDataQuery.Data?> = conferenceFlow().flatMapLatest {
```
martinbonnin, line 48: Small comment: I'd like this to take a "conference" parameter at some point `fun conferenceDataFlow(conference: String)`. I like the idea that a given "screen" listens to only one conference. In the current state of affairs, something could call `setConference(newconference)` under the hood and change all the displayed data.

It's definitely not related to this PR, just thinking about how we want to model things. In all cases, androidx datastore looks good to me 👍 (not to be confused with the backend GCP datastore 😅 )

martinbonnin (approved):

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
yschimke: I think it's the one below.  I can work harder to remove this form of it.  It should clean up the app state.

yschimke:
Going to split this up into incremental PRs

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
yschimke: I'm working towards this in https://github.com/joreilly/Confetti/pull/257

PR #255, opened 2023-03-05
Use FlowSettings

yschimke:
Use FlowSettings.

This also exposes some hidden blocking calls, in this PR these are TODOs. A bigger PR https://github.com/joreilly/Confetti/pull/245 I'm splitting avoid this.


yschimke:
@joreilly This is ready now. By the end of plan in https://github.com/joreilly/Confetti/pull/257, it should be clean again.

PR #256, opened 2023-03-05
Avoid debug firebase dependency

yschimke:
Don't assume Firebase is setup in debug builds


martinbonnin:
No committing of the google-services.json then?

yschimke:
@martinbonnin This seemed like a good change anyway. Not a vote either way.

martinbonnin:
👍 Asking cause I'm starting to think about Bookmarks that will require Firebase auth at some point. I might commit a google-services.json at that point so that someone cloning the repo would still be able to use the feature.

yschimke:
What about a test only instance, leave the production one separate?

martinbonnin:
That should work 👍

PR #257, opened 2023-03-05
Split up SessionDetail ViewModel to use specific query.

yschimke:
Splitting up ConfettiViewModel, use targetted specific queries.

This isn't very simple yet. Plan

1) Demonstrate this approach incrementally on Wear
a) SessionDetails (this)
b) Session
c) Conference Day
d) Conference

2) Apply to android also

3) Delete ConfettiViewModel, ensure clean navigation with conference in all paths.

Need to land after https://github.com/joreilly/Confetti/pull/255

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
@@ -1,5 +1,6 @@
 package dev.johnoreilly.confetti.di
 
+import com.apollographql.apollo3.cache.normalized.FetchPolicy
 import dev.johnoreilly.confetti.ConfettiViewModel
 import dev.johnoreilly.confetti.sessiondetails.SessionDetailsViewModel
 import dev.johnoreilly.confetti.speakerdetails.SpeakerDetailsViewModel
@@ -10,5 +11,9 @@ val appModule = module {
     viewModel { ConfettiViewModel() }
     viewModel { SessionDetailsViewModel(get(), get()) }
     viewModel { SpeakerDetailsViewModel(get(), get()) }
-
+    single {
+        // Assume an online first strategy for Mobile
+        // But use Cache for initial results
+        FetchPolicy.CacheAndNetwork
+    }
 }
--- a/shared/build.gradle.kts
+++ b/shared/build.gradle.kts
@@ -44,7 +44,7 @@ kotlin {
                 api(libs.koin.core)
 
                 api(libs.apollo.runtime)
-                implementation(libs.bundles.apollo)
+                api(libs.bundles.apollo)
             }
         }
         val commonTest by getting {
--- a/shared/src/commonMain/graphql/Queries.graphql
+++ b/shared/src/commonMain/graphql/Queries.graphql
@@ -14,6 +14,7 @@ query GetConferenceData($first: Int! = 1000, $after: String = null) {
         ...RoomDetails
     }
     config {
+        id
         name
         timezone
     }
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
@@ -1,0 +1,42 @@
+package dev.johnoreilly.confetti
+
+import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.cache.normalized.api.MemoryCacheFactory
+import com.apollographql.apollo3.cache.normalized.normalizedCache
+import com.apollographql.apollo3.cache.normalized.sql.SqlNormalizedCacheFactory
+import dev.johnoreilly.confetti.di.getDatabaseName
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+
+class ApolloClientCache : KoinComponent {
+    val _clients = mutableMapOf<String, ApolloClient>()
+    val mutex = Mutex(false)
+
+    suspend fun getClient(conference: String): ApolloClient {
+        return mutex.withLock {
+            _clients.getOrPut(conference) {
+                clientFor(conference, conference != "all")
+            }
+        }
+    }
+
+    private fun clientFor(
+        conference: String,
+        writeToCacheAsynchronously: Boolean
+    ): ApolloClient {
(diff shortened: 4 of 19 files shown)


joreilly:
What's the reason again for different fetch policy for Android and Wear?  Just wondering, if they were the same, if we could keep config in shared code.

joreilly:
btw looks like issue running JVM `main` function.

```Exception in thread "main" org.koin.core.error.InstanceCreationException: Could not create instance for '[Singleton:'dev.johnoreilly.confetti.ConfettiRepository']'
	at org.koin.core.instance.InstanceFactory.create(InstanceFactory.kt:59)
	at org.koin.core.instance.SingleInstanceFactory.create(SingleInstanceFactory.kt:46)
	at org.koin.core.instance.SingleInstanceFactory$get$1.invoke(SingleInstanceFactory.kt:53)
	at org.koin.core.instance.SingleInstanceFactory$get$1.invoke(SingleInstanceFactory.kt:51)
	at org.koin.mp.KoinPlatformTools.synchronized(KoinPlatformTools.kt:20)
	at org.koin.core.instance.SingleInstanceFactory.get(SingleInstanceFactory.kt:51)
	at org.koin.core.registry.InstanceRegistry.resolveInstance$koin_core(InstanceRegistry.kt:106)
	at org.koin.core.scope.Scope.resolveValue(Scope.kt:246)
	at org.koin.core.scope.Scope.resolveInstance(Scope.kt:231)
	at org.koin.core.scope.Scope.get(Scope.kt:210)
	at MainKt.main(Main.kt:20)
	at MainKt$main$3.invoke(Main.kt)
	at MainKt$main$3.invoke(Main.kt)
	at kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt$createCoroutineUnintercepted$$inlined$createCoroutineFromSuspendFunction$IntrinsicsKt__IntrinsicsJvmKt$1.invokeSuspend(IntrinsicsJvm.kt:205)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlin.coroutines.ContinuationKt.startCoroutine(Continuation.kt:115)
	at kotlin.coroutines.jvm.internal.RunSuspendKt.runSuspend(RunSuspend.kt:19)
	at MainKt.main(Main.kt)
Caused by: org.koin.core.error.NoBeanDefFoundException: No definition found for class:'com.apollographql.apollo3.cache.normalized.FetchPolicy' q:''. Check your definitions!
	at org.koin.core.scope.Scope.throwDefinitionNotFound(Scope.kt:298)
	at org.koin.core.scope.Scope.resolveValue(Scope.kt:268)
	at org.koin.core.scope.Scope.resolveInstance(Scope.kt:231)
	at org.koin.core.scope.Scope.get(Scope.kt:210)
	at dev.johnoreilly.confetti.di.KoinKt$commonModule$1$1.invoke(Koin.kt:34)
	at dev.johnoreilly.confetti.di.KoinKt$commonModule$1$1.invoke(Koin.kt:23)
	at org.koin.core.instance.InstanceFactory.create(InstanceFactory.kt:52)
```

joreilly:
Yeah, and I should have had CI task for JVM stuff :)

yschimke:
> What's the reason again for different fetch policy for Android and Wear? Just wondering, if they were the same, if we could keep config in shared code.

I think Wear has to avoid making a network request on every navigation. It's ok to assume on mobile, that requests are almost free.

I think we should end up with an offline-first app https://developer.android.com/topic/architecture/data-layer/offline-first

And look at some approach with firebase notifications when data does change. Or for Wear, using the Mobile as an indicator that data is stale.

martinbonnin:
> I think Wear has to avoid making a network request on every navigation. It's ok to assume on mobile, that requests are almost free.

On mobile, the data for a conference is always all loaded up at startup. IIRC there are 2 network request:

1. GetConferences: all available conferences (very small, refreshed in the background)
2. GetConferenceData(conference): all data for a conference (bigger, 200k json again IIRC)

> I think we should end up with an offline-first app

The GraphQL cache does that mostly for free (writes are online only). This is the reason for `CacheAndNetwork`. The cache is displayed immediatly. If there is some network, the data is updated in the background.

> And look at some approach with firebase notifications when data does change. Or for Wear, using the Mobile as an indicator that data is stale.

What's cheaper between using Wifi vs getting the data from the Mobile (using bluetooth I guess?)? If Bluetooth is cheaper we might even use the mobile as source of truth for the data?

joreilly:
>I think Wear has to avoid making a network request on every navigation. It's ok to assume on mobile, that requests are almost free.

Actually, now that I think about it I thought we were offline-first for mobile as well.  Hmm, though I guess we're also attempting to make network request as well....at least on initial load and when pull to refresh triggers `refresh()` call.  The other place I guess we're always hitting network for is loading list of conferences.....  I don't think it should happen though for any other navigation?

yschimke:
Added a test in this PR.  Also changed the main to only consume the first result.

joreilly:
ah, sorry I missed @martinbonnin 's reply

yschimke:
@joreilly Yep, maybe as part of my change with different queries.

I can revert this part on Wear and do it separately once all the queries are in place.

yschimke:
Happy to take guidance. My aim would still be that as part of opening and closing the Wear app a few times, we generally don't hit the network at all.  Even for an optimistic refresh.

yschimke:
@martinbonnin Will the current top level refresh in ConfettiViewModel, also populate the cache for individual queries?

So maybe 

    GetConferences: all available conferences (very small, refreshed in the background) - refreshed on a scheduled daily job?
    GetConferenceData(conference): all data for a conference (bigger, 200k json again IIRC) - `CacheAndNetwork`
    GetSpecificData() - assume this is already cache by GetConferenceData, so just `Cache`.
    
Anything refreshing from ViewModel doesn't work well for Wear Complication, Tiles or Mobile Widgets.

joreilly:
I guess for Wear app we'd at least still have the refresh button to pull down latest data from network.

martinbonnin:
Maybe stupid question but how about we pass the fetch policy up to the UI?  

```
GetConferenceData(conference, fetchPolicy)
```

The UI knows whether it wants to be updated in the background or not. Also looks like we'll want WearOS to do `NetworkOnly` on some external signal (notification or phone push)

If that doesn't work then multiple ViewModel methods work too.

joreilly:
I'll merge this now anyway to least have core changes in place and we can, as ever, follow up if needed if we want to change approach to anything.

PR #261, opened 2023-03-05
Split up Sessions Screen ViewModel.

yschimke:
Step 2.

I suspect caching for this won't make sense until I stop using ConfettiViewModel with it's uber query, and change to a refresh that queries Sessions, Rooms and Speakers in matching queries.

But pulling a thread and seeing where it goes.

### DIFF
--- a/shared/src/commonMain/graphql/Queries.graphql
+++ b/shared/src/commonMain/graphql/Queries.graphql
@@ -20,6 +20,20 @@ query GetConferenceData($first: Int! = 1000, $after: String = null) {
     }
 }
 
+query GetSessions($first: Int! = 1000, $after: String = null) {
+    sessions(first: $first, after: $after) {
+        nodes {
+            ...SessionDetails
+        }
+        pageInfo {
+            endCursor
+        }
+    }
+    config {
+        timezone
+    }
+}
+
 query GetSession($id: String!){
 	session(id: $id) {
 	    ...SessionDetails
--- a/wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessions/SessionsViewModel.kt
+++ b/wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessions/SessionsViewModel.kt
@@ -1,0 +1,40 @@
+package dev.johnoreilly.confetti.wear.sessions
+
+import androidx.lifecycle.SavedStateHandle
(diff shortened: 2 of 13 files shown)


shared/src/commonMain/graphql/Queries.graphql
```diff
@@ -20,6 +20,20 @@ query GetConferenceData($first: Int! = 1000, $after: String = null) {
     }
 }
 
+query GetSessions($first: Int! = 1000, $after: String = null) {
```
yschimke, line 23: Seems a shame we can't query for session on a day.  Should I change the query?

wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessions/SessionsViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+) : ViewModel() {
+    private val conferenceDay: ConferenceDayKey =
+        SessionsDestination.fromNavArgs(savedStateHandle)
+
+    val session: StateFlow<List<SessionDetails>?> = flow {
+        val resultsFlow = repository.sessions(conferenceDay.conference)
+
+        val sessions = resultsFlow
+            .map {
+                val timeZone = TimeZone.of(it.dataAssertNoErrors.config.timezone!!)
```
yschimke, line 31: Any guidance on dataAssertNoErrors?

I suspect I should have a ScreenState with Loading, Successful and Failed.

shared/src/commonMain/graphql/Queries.graphql
martinbonnin: The API doesn't allow this. Can you open an issue? I'll add this.

martinbonnin: Feel free to specify how you'd want this to work i.e:

* `sessions(after: "2023-04-13-00:00", before: "2023-04-14-00:00")`
* `sessions(day: "2023-04-13")`
* something else?

wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessions/SessionsViewModel.kt
yschimke: @joreilly I'll follow up on this, would rather complete the navigation changes first.

PR #263, opened 2023-03-09
Make Home a non special destination

yschimke:
Completing refactoring the app navigation.

Note: there are still TODOs where we were already having blocking calls in the main path.

Next, probably profile and optimise queries to make sure we don't over fetch.

### DIFF
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
@@ -1,5 +1,6 @@
 package dev.johnoreilly.confetti
 
+import com.apollographql.apollo3.ApolloCall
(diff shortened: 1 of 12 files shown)


yschimke:
@martinbonnin Am I misunderstanding cache normalisation? If I call `GetSessionsQuery`, shouldn't that cache all instances so I'll get a cache results for `GetSessionQuery(sessionId)`

https://www.apollographql.com/blog/apollo-client/caching/demystifying-cache-normalization/

martinbonnin:
@yschimke you got it right, it "just" requires some configuration because the GraphQL spec doesn't say anything about caching or "ids" so we have to tell the runtime what fields are ids. I opened https://github.com/joreilly/Confetti/pull/264 for that. If you're curious, there's also some doc there: https://www.apollographql.com/docs/kotlin/caching/declarative-ids/

PR #264, opened 2023-03-09
Configure cache ids

martinbonnin:
See https://github.com/joreilly/Confetti/pull/263


yschimke:
This is pretty awesome. I mean I'd prefer automatic, but it keeps it a nice and clean request flow.

martinbonnin:
> I mean I'd prefer automatic

Me too! We should create the graphql-cache-working-group!

PR #277, opened 2023-03-11
Introduce UI State models

yschimke:
Avoid using `dataAssertNoErrors`. Instead have a model with Loading, Success, Error.

Follow up work to improve errors and loading.

### DIFF
--- a/shared/src/commonMain/graphql/Queries.graphql
+++ b/shared/src/commonMain/graphql/Queries.graphql
@@ -17,6 +17,7 @@ query GetConferenceData($first: Int! = 1000, $after: String = null) {
         id
         name
         timezone
+        days
     }
 }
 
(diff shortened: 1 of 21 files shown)


martinbonnin:
Very cool 👍 

> Avoid using `dataAssertNoErrors`

FWIW, I'm not super happy with this `dataAssertNoErrors` and with error handling in general. 
It's not easy because of partial data and stuff like "GraphQL errors returning 200" but we put some thoughts over there: https://github.com/apollographql/apollo-kotlin/issues/4711

yschimke:
+1 to making errors just another state.  Not sure what I want, but some random thoughts.

I'm not sure what I want, but I think something like https://github.com/skydoves/sandwich, with an additional Loading state would work really well for Compose screens.

We found this section abstraction for screens useful. 

https://github.com/google/horologist/blob/9cb3879ed6b044f702b2b2ad0de06f92e5985a09/sample/src/main/java/com/google/android/horologist/sectionedlist/stateful/SectionedListStatefulScreen.kt#L118-L151

Might be useful if you a screen built from multiple queries, each of which has Loading/Success/Failed states.

PR #284, opened 2023-03-12
Add bookmarks

martinbonnin:
Add bookmarks using Firebase auth and Google cloud datastore and the same GraphQL backend

It's working only with Google signin so far. We should certainly add email/apple too.


joreilly:
Very cool!!!

martinbonnin:
Good catch from @joreilly to reuse the existing Firebase project instead of creating a new one. I'll do that in a few

martinbonnin:
Switched the Firebase token verification and authentication Firebase project we already have (linked to Google Cloud project confetti-164da). Note the GraphQL API and landing page run on a different GCP project: confetti-349319

The oauth screen is being verified. In the meantime, you can test if you use the same Google adress as the one that's linked to firebase.

Also random Sunday night thoughts: I'm not sure where to hoist the User state. I'll read https://developer.android.com/jetpack/compose/state-hoisting and certainly make a new PR. Also, we'll certainly want to make the current logged in user part of the state somehow so that we can recompose accordingly. But that's also something for another PR :)

PR #295, opened 2023-03-14
Adding GoogleSignIn for Wear

yschimke:


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -8,7 +8,7 @@ object AndroidSdk {
 
 object WearSdk {
     const val min = 26
-    const val compile = 33
+    const val compile = "UpsideDownCake"
```
martinbonnin, line 11: 😋

yschimke:
cc @luizgrp Any feedback?

yschimke:
Some thoughts

1) It's worth comparing the implementation effort for Mobile, although that uses Firebase auth

https://github.com/joreilly/Confetti/blob/27d1b958f2518ffcf9210ee7dc25f483fae5a7c9/androidApp/src/main/java/dev/johnoreilly/confetti/account/Authentication.kt

2) We should consider just using token sharing, so the mobile drives everything here normally

https://google.github.io/horologist/auth-tokenshare-guide/

wearApp/src/main/java/dev/johnoreilly/confetti/wear/settings/SettingsListView.kt
```diff
@@ -36,7 +40,25 @@ fun SettingsListView(
         }
 
         item {
-            StandardChip(label = "Change Conference", onClick = conferenceCleared)
+            StandardChip(
+                label = "Change Conference",
```
luizgrp, line 44: nit: add texts from this file to strings resources file

wearApp/src/main/java/dev/johnoreilly/confetti/wear/auth/GoogleSignInPromptScreen.kt
```diff
@@ -0,0 +1,74 @@
...
+            SignInChip(
+                onClick = {
+                    navigateToGoogleSignIn()
+                },
+                chipType = StandardChipType.Secondary
+            )
+        }
+    }
+
+    if (showAlreadySignedInDialog) {
```
luizgrp, line 60: the logic [here](https://github.com/joreilly/Confetti/pull/295/files#diff-2e42dcaf3bfd1f6ea686706be36118e3da79a5879291bbd8761fc22254c5028bR51) seems to make sure this screen is not called when it's already signed, I guess it's just being cautious here?

joreilly:
On 2 above, I believe you can use Wear app on it's own (without mobile app installed)?

luizgrp (commented):
> Any feedback?

@yschimke no, just a question!

thanks for trying the lib, is this the PR that made you raise https://github.com/google/horologist/issues/1106? if so, could you please details which extra dependencies had to be pulled? in this PR I only see the auth ones being added plus "kotlinx-coroutines-play-services" and potentially "play-services-auth"?

wearApp/src/main/java/dev/johnoreilly/confetti/wear/auth/GoogleSignInPromptScreen.kt
yschimke: Yeah, I think this screen should work even if we decided to do this at startup.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/settings/SettingsListView.kt
yschimke: Yep, I think that's a wider problem.

yschimke:
> > Any feedback?
> 
> @yschimke no, just a question!
> 
> thanks for trying the lib, is this the PR that made you raise [google/horologist#1106](https://github.com/google/horologist/issues/1106)? if so, could you please details which extra dependencies had to be pulled? in this PR I only see the auth ones being added plus "kotlinx-coroutines-play-services" and potentially "play-services-auth"?

Yes - in the end all dependencies are justifiable to put into the app dependencies, but there are certainly some that are part of the public API that we are not marking as such.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/auth/GoogleSignInPromptScreen.kt
luizgrp: In the startup I'd say it would be better to just dismiss the screen instead of blocking the user journey with a dialog, but yeah, other scenarios might be appropriate to display the dialog to not look like the button is broken 👍

martinbonnin:
Quick question: if this is using plain Google Signin, I guess it doesn't share the uid with Firebase, right? So if you login on your watch and on your phone using the same Google account, you won't have the same bookmarks? 

> We should consider just using token sharing, so the mobile drives everything here normally

I was thinking of adding Apple signin to the mobile app. I think this is a requirement for the appstore. And email too as the fallback option. Not sure how that would work on Wear OS. 

All in all, sounds like token sharing is the way?

yschimke:
> I was thinking of adding Apple signin to the mobile app. I think this is a requirement for the appstore. And email too as the fallback option. Not sure how that would work on Wear OS.
> 
> All in all, sounds like token sharing is the way?

Let's land this, and I can follow up bridging to Firebase auth so the ids match.  I can even look to contribute that back upstream.

We could also use the email from the token? If we are authed to the server.

martinbonnin:
> Let's land this, and I can follow up bridging to Firebase auth so the ids match. I can even look to contribute that back upstream.

👍 

> We could also use the email from the token? If we are authed to the server.

Not sure what you mean. The GraphQL bookmarks use the Firebase `uid` for auth. email is asked but never used

yschimke:
> Not sure what you mean. The GraphQL bookmarks use the Firebase `uid` for auth. email is asked but never used

If the server has an authed token including email, it could use the email instead.

But it sounds like the simpler fix is to use Firebase auth on wear if possible.

martinbonnin:
> If the server has an authed token including email, it could use the email instead.

I'd like to avoid that if possible. `uid` sounds like a more robust identifier. Also I think it'd allow stuff like promoting anonymous users into "real" ones, etc...

joreilly:
@yschimke did you want to make more changes to this PR or are you happy to merge for now and follow up with other stuff/

yschimke:
> @yschimke did you want to make more changes to this PR or are you happy to merge for now and follow up with other stuff?

It's functional, so merge now.  I'll follow up with further changes, but likely involve exploration.

Generally I'll use draft to signal when I'm planning further changes. So non-draft, CI is passing, and no blocking comments is fine to land.

Issue #268, opened 2023-03-10
API: add "views" for sessions


martinbonnin:
It'd be nice to separate the actual session storage from the session "views" so that:

1. the data from sessions that are not scheduled yet is still reachable through the API
2. the logic to display an agenda is factored in the backend (in a way, this is server side rendering)

Especially for 2D agendas (grid), there's usually a log of business logic (using a linear time scale is usually not very nice, lunch breaks, etc... need special handling, etc...). We could model all of that in the schema.

```graphql 
interface Slot2d { x y w h }

type BreakSlot2d: Slot2d
type LunchSlot2d: Slot2d
type SessionSlot2d: Slot2d
# ...

{
  agenda2d {
    headers {
      name
    }
    slots {
      x
      y
      w
      h
      ... on SessionSlot {
        title
        start
        speakers { 
          avatarUrl
        }
      }
    }
  }
}

```

PR #297, opened 2023-03-15
Remove warnings, use kotlinx.datetime everywhere

martinbonnin:


buildSrc/src/main/kotlin/options.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.jvm.toolchain.JavaLanguageVersion
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmOptions
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+
+fun Project.configureCompilerOptions(jvmVersion: Int = 8) {
+    tasks.withType(KotlinCompile::class.java).configureEach {
+        it.compilerOptions {
+            //allWarningsAsErrors.set(true)
```
martinbonnin, line 12: I commented this for now as it's a bit painful during development (unused variables, etc...) I'll enable it from time to time keep warnings in check

shared/src/commonMain/graphql/schema.graphqls
```diff
@@ -86,7 +86,7 @@ type Session {
   """
   language: String
 
-  room: Room @deprecated(reason: "use rooms instead")
+  room: Room
```
martinbonnin, line 89: I "undeprecated" 😅  `session.room` because it was easy and `rooms` only make sense for service sessions that would somewhat be better modeled using https://github.com/joreilly/Confetti/issues/268. 
Ultimately, I think `@room` needs to go completely and be moved to a `SessionSlot` model instead.

joreilly:
nice!

martinbonnin:
@yschimke can you take a look? It touches a non-trivial amount of wear os code (all the Java time stuff).

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -108,6 +108,7 @@ class ConfettiRepository(
                 // handle network failures by swallowing the error
             }
             .mapNotNull {
+                println("conferences: ${it.data?.conferences}")
```
yschimke, line 111: deliberate?

martinbonnin: Absolutely not, thanks 🙏

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
```diff
@@ -24,7 +24,7 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        val startingConferenceAsync = lifecycleScope.async {
+        lifecycleScope.async {
```
yschimke, line 27: I guess we can remove this now? I can do it later.

yschimke (approved):
Thanks for the cleanup

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
martinbonnin: It was triggering a warning, feel free to remove 👍

PR #298, opened 2023-03-15
wide screen updates

joreilly:
Screenshot_20230315_140839


joreilly:
This needs further cleanup but good start hopefully for now.

PR #301, opened 2023-03-15
Make "conference" part of the deeplinks

martinbonnin:
This PR:

* adds the `sessions_route/{conference}` route
* puts the `conferences_route` route in the navgraph so it's also reachable through deep links
* adds basic error handling

Heavily inspired by @yschimke work on the Wear OS app 🙏


androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiApp.kt
```diff
@@ -24,80 +24,151 @@ import androidx.compose.material3.Scaffold
...
-                            .consumeWindowInsets(padding)
+    val repository: ConfettiRepository = get()
+    val conference by repository.getConferenceFlow()
+        .collectAsState(initial = null)
+
+    if (conference == null) {
+        // Reading from the settings
+        CircularProgressIndicator()
+    } else {
+        val initialRoute = if (conference == CONFERENCE_NOT_SET) {
```
yschimke, line 74: I'll try to dig it out, but I'm pretty sure I've seen advice to not change the startDestination in this case. But if it's working, let's go with it until I can find that and it's worth changing.

yschimke (approved):
LGTM

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiApp.kt
martinbonnin: Yea, it's recomposing the whole `NavHost`... Feels a bit weird but on the other hand, it's almost as if we restarted the app. I'm curious if you find anything.

androidApp/src/main/java/dev/johnoreilly/confetti/navigation/ConfettiNavHost.kt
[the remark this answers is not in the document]
martinbonnin: It's all in `ConfettiApp` now. It maps quite nicely with the Wear OS app.

joreilly:
Nice changes!

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/navigation/SessionsDestination.kt
```diff
@@ -18,15 +26,25 @@ fun NavGraphBuilder.sessionsGraph(
...
     onSwitchConferenceSelected: () -> Unit,
+    conference: String,
 ) {
-    composable(route = SessionsDestination.route) {
+    composable(
+        route = SessionsDestination.route,
+        arguments = listOf(
+            navArgument(SessionsDestination.conferenceArg) {
+                type = NavType.StringType
+                defaultValue = "cds"
```
martinbonnin, line 36:
```suggestion
                defaultValue = "conference"
```

PR #309, opened 2023-03-16
Fix Wear app for play store

yschimke:


yschimke:
If change is bigger, I'll have to fix up tonight.

PR #327, opened 2023-03-16
Use gradle-build-action

yschimke:
Nice speed up through caching the jobs (when run on main).

### DIFF
--- a/.github/workflows/android_wear_jvm.yml
+++ b/.github/workflows/android_wear_jvm.yml
@@ -11,10 +11,17 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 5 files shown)


martinbonnin:
We'll want `gradle-home-cache-cleanup: true`:

```
        with:
          gradle-home-cache-cleanup: true
``` 


https://github.com/gradle/gradle-build-action#removing-unused-files-from-gradle-user-home-before-saving-to-cache

joreilly:
>We'll want gradle-home-cache-cleanup: true:

@yschimke is that something that has to be setup somewhere?

martinbonnin:
I'll do it in a follow up PR

PR #328, opened 2023-03-16
Wear instrumentation test

yschimke:
Add a Wear instrumentation test, guarded by a label so it doesn't burn github CI minutes.

### DIFF
--- a/.github/workflows/wear_instrumentation.yml
+++ b/.github/workflows/wear_instrumentation.yml
@@ -1,4 +1,4 @@
-name: Android, Wear and JVM CI
+name: Wear Instrumentation
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/wear_instrumentation.yml
```diff
@@ -0,0 +1,65 @@
...
+# Cancel any current or previous job from the same PR
+concurrency:
+  group: wear-instrumentation-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: macos-latest
+    # TODO change to a test specific label
+    #if: contains(github.event.pull_request.labels.*.name, 'enhancement')
```
yschimke, line 16: @joreilly  can you grant me enough rights to set labels, no point running these on every build, but selectively, and maybe daily/weekly?

yschimke: And maybe make a "test" label?

joreilly: @yschimke I just added you as a collaborator to the project (had thought I'd done that before).  I also created "test" label

PR #339, opened 2023-03-17
CI: opt-in gradle-home-cache-cleanup: true

martinbonnin:
I'm not 100% clear where the good place to put this is. Looks like the `gradle-build-action` config and cache key is global to the workflow so putting it in the initial setup is the correct place but maybe it needs to be in the last one, not 100% sure.

See also https://github.com/gradle/gradle-build-action/issues/630

### DIFF
--- a/.github/workflows/build-apps.yml
+++ b/.github/workflows/build-apps.yml
@@ -22,6 +22,8 @@ jobs:
 
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
 
       - name: Build
         uses: gradle/gradle-build-action@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/build-apps.yml
```diff
@@ -22,6 +22,8 @@ jobs:
 
       - name: Setup Gradle
         uses: gradle/gradle-build-action@v2
+        with:
+          gradle-home-cache-cleanup: true
```
yschimke, line 26: I think this is the only one that actually runs on main.

martinbonnin: Makes sense. That also means if a PR workflow uses a big dependency, the next MAIN workflow will most likely delete it because it wasn't used during the run. I think that's acceptable as MAIN should do most of the work. Anyways, I'll remove the other ones.

PR #352, opened 2023-03-19
Add Wear pairing flow and settings sync.

yschimke:
Use Horologist helpers for data syncing, including installing the wear app.  Then share the theme as proto.

To follow up:

- Some screens are not as readable, see video.

[weartheme.webm](https://user-images.githubusercontent.com/231923/226165913-244bfecd-8aea-4d55-a7d7-44a357861439.webm)


yschimke:
@joreilly any idea about the build failure? How do I have common wire proto shared? If I run the proto step manually, then it works from then on.

I suspect Wire doesn't handle KMM perfectly yet.

joreilly:
Hmm, not sure tbh.  @martinbonnin in case you've come across this?  Maybe some expect/actual setup needed to handle this.

martinbonnin:
Ping @oldergod. Is KMM supported with Wire? Or do we need to wire (pun intended) the task manually?

yschimke:
The tasks for wire look inconsistent. I might try moving to commonMain

oldergod:
Currently, only protos in `commonMain` will be detected yes.

androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
```diff
@@ -18,29 +18,24 @@ import dev.johnoreilly.confetti.speakerdetails.SpeakerDetailsViewModel
...
-        Authentication()
-    }
-    single<TokenProvider> {
-        object : TokenProvider {
-            override suspend fun token(forceRefresh: Boolean): String? {
-                return get<Authentication>().idToken(forceRefresh)
-            }
-        }
-    }
+    singleOf(::WorkManagerConferenceRefresh).withOptions { bind<ConferenceRefresh>() }
```
yschimke, line 36: Ooops

oldergod:
And I don't think Wire is gonna support other platforms for we'd have to generate expects/actuals. Maybe we could detect protos in `<platform>Main` and generate it all in `commonMain`, I don't know.

yschimke:
@oldergod should I put this in a non KMP target?  Still not finding things in commonMain/proto

oldergod:
There's a current bug which might be the cause https://github.com/square/wire/issues/2411
I'm in the middle of a PR which should improve everything https://github.com/square/wire/pull/2412 which I hope to land very soon™

yschimke:
@oldergod Thanks, that was it.  Slightly different fix.

```
// https://github.com/square/wire/issues/2411
val wireTask = tasks.withType<com.squareup.wire.gradle.WireTask>()
tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
    dependsOn(wireTask)
}
afterEvaluate {
    wireTask.forEach {
        if ("generateCommon.+MainProtos".toRegex().matches(it.name)) {
            it.enabled = false
        }
    }
}
```

yschimke:
cc @garanj

PR #356, opened 2023-03-19
Make the initial loading screen part of the nav graph

martinbonnin:
This PR: 

* Uses some of the concepts from https://github.com/android/nowinandroid/pull/347/commits/d108459bda50d84b383072b0fe0d9d670729775f 
  * no more ConfettiDestination abstract class
  * instead each deeplink gets its own "Key", like on Wear OS
  * In NIA, they're called Args but it's the same pattern
  * TBD if we want those "type safe deeplinks" part of "shared" or not
* Puts `SpeakersDetailsViewModel` in shared if we ever want to use it from iOS
* Adds an optional "splash screen" destination. The "splash screen" is responsible for loading the current conference setting. It's very fast but if you open the app from a deeeplink that contains "conference" already, it'll skip that step completely.


androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
```diff
@@ -19,13 +19,15 @@ import androidx.compose.ui.platform.LocalContext
...
 import dev.johnoreilly.confetti.fragment.SessionDetails
-import dev.johnoreilly.confetti.ui.ConfettiScaffold
 import org.koin.androidx.compose.getViewModel
 
 @Composable
-fun SessionDetailsRoute(onBackClick: () -> Unit, viewModel: SessionDetailsViewModel = getViewModel()) {
+fun SessionDetailsRoute(conference: String, sessionId: String, onBackClick: () -> Unit) {
+    val viewModel: SessionDetailsViewModel = getViewModel<SessionDetailsViewModel>().apply {
+        configure(conference, sessionId)
+    }
```
martinbonnin, line 30: I'm not specifically happy with this pattern of passing arguments to a ViewModel. Ideally, we could compute those from `SavedStateHandle` directly in the VM but that won't work on iOS :(.

joreilly:
Looks cleaner

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/AppViewModel.kt
```diff
@@ -3,24 +3,25 @@ package dev.johnoreilly.confetti
...
-            SharingStarted.WhileSubscribed(),
-            null
-        )
-
-    suspend fun setConference(conference: String) {
-        repository.setConference(conference)
-    }
+    val conference: StateFlow<String?> = repository
+        .getConferenceFlow()
+        .onStart {
```
yschimke, line 19: Intended?

yschimke (commented):
Does it fit with this guidance? https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/AppViewModel.kt
martinbonnin: Nope, thanks!

martinbonnin:
> Does it fit with this guidance? https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

I'm pretty sure not, this touches the code once the app is running, it doesn't touch the Theme or anything like this

androidApp/src/main/java/dev/johnoreilly/confetti/splash/SplashView.kt
```diff
@@ -0,0 +1,26 @@
...
+fun SplashRoute(
+    navigateToConferences: (ConferencesKey) -> Unit,
+    navigateToSessions: (SessionsKey) -> Unit
+) {
+    val viewModel = getViewModel<AppViewModel>()
+
+    val conference by viewModel.conference.collectAsStateWithLifecycle()
+    when (val conference1 = conference) {
+        null -> LoadingView()
+        CONFERENCE_NOT_SET -> navigateToConferences(ConferencesKey)
```
yschimke, line 23: Should these be in a `SideEffect {}`?

martinbonnin: Yea, that felt a bit weird. I guess so? It's working though

yschimke:
> > Does it fit with this guidance? https://developer.android.com/develop/ui/views/launch/splash-screen/migrate
> 
> I'm pretty sure not, this touches the code once the app is running, it doesn't touch the Theme or anything like this

No objection from me, it's an awkward problem.

Another option using the splashscreen API - https://developer.android.com/develop/ui/views/launch/splash-screen#suspend-drawing

androidApp/src/main/java/dev/johnoreilly/confetti/splash/SplashView.kt
martinbonnin: Updated to use `SideEffect {}`

martinbonnin:
> Another option using the splashscreen API 

I named this PR wrong, it's not about a SplashScreen as per https://developer.android.com/develop/ui/views/launch/splash-screen but more initial loading screen, when the app starts to draw. I'll rename things.

androidApp/src/main/java/dev/johnoreilly/confetti/splash/SplashView.kt
```diff
@@ -20,7 +21,11 @@ fun SplashRoute(
     val conference by viewModel.conference.collectAsStateWithLifecycle()
     when (val conference1 = conference) {
         null -> LoadingView()
-        CONFERENCE_NOT_SET -> navigateToConferences(ConferencesKey)
+        CONFERENCE_NOT_SET -> {
+            SideEffect {
+                navigateToConferences(ConferencesKey)
+            }
+        }
         else -> navigateToSessions(SessionsKey(conference1))
```
yschimke, line 29: Probably also for navigateToSessions. 

Generally my mental model is SideEffect can run after the UI is drawn. And tell the compose runtime, this shouldn't be skipped.

martinbonnin: Makes sense 👍

yschimke:
> > Another option using the splashscreen API
> 
> I named this PR wrong, it's not about a SplashScreen as per https://developer.android.com/develop/ui/views/launch/splash-screen but more initial loading screen, when the app starts to draw. I'll rename things.

Sure, but I think dragging out the splashscreen slightly, might be better than

1) Splashscreen
2) Loading Indicator (avoid this)
3) Content

martinbonnin:
Agreed, will do in a separate PR

PR #357, opened 2023-03-19
Collapse top-bar while scrolling

marcellogalhardo:
The default behaviour is to keep top bar always visible. However, if NavRail is not visible (i.e., phones), we can collapse the top bar while scrolling to save visible space and let the user focus on the screen's content.

| TopBar Expanded | TopBar Collapsed |
|---|---|
| Screenshot_1679245606 | Screenshot_1679245614 |
| Screenshot_1679245619 | Screenshot_1679245623 |


martinbonnin (approved):

joreilly:
Love how more straightforward stuff like this has become with Compose.

PR #359, opened 2023-03-19
Format date tabs

marcellogalhardo:
* Instead of printing the `LocalDate` without any format, fallbacks to use the `FormatStyle.MEDIUM`.

Screenshot_1679247734

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListView.kt
@@ -41,7 +41,6 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import dev.johnoreilly.confetti.SessionsUiState
-import dev.johnoreilly.confetti.SpeakersUiState
 import dev.johnoreilly.confetti.account.Authentication
 import dev.johnoreilly.confetti.fragment.SessionDetails
 import dev.johnoreilly.confetti.isBreak
@@ -50,9 +49,11 @@ import dev.johnoreilly.confetti.ui.ErrorView
 import dev.johnoreilly.confetti.ui.LoadingView
 import dev.johnoreilly.confetti.ui.component.ConfettiTab
 import dev.johnoreilly.confetti.ui.component.pagerTabIndicatorOffset
+import dev.johnoreilly.confetti.utils.format
 import kotlinx.coroutines.launch
 import org.koin.androidx.compose.get
-
+import java.time.format.DateTimeFormatter
+import java.time.format.FormatStyle
 
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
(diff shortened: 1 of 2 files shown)


androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListView.kt
```diff
@@ -129,6 +130,7 @@ fun SessionListView(
     }
 }
 
+private val dateFormatter = DateTimeFormatter.ofLocalizedDate(FormatStyle.MEDIUM)
```
marcellogalhardo, line 133: I didn't add it inside Koin as we may want to use different formats for different screens - and it seems an overkill to use qualifiers and an`get<>()` inside the composable for that.

If you prefer to move there, let me know.

androidApp/src/main/java/dev/johnoreilly/confetti/utils/LocalDateTime.kt
```diff
@@ -0,0 +1,9 @@
+package dev.johnoreilly.confetti.utils
+
+import kotlinx.datetime.LocalDate
+import kotlinx.datetime.toJavaLocalDate
+import java.time.format.DateTimeFormatter
+
+/** @see [DateTimeFormatter.format] */
+fun DateTimeFormatter.format(date: LocalDate): String =
```
marcellogalhardo, line 8: I have tried to place it inside `shared` but I couldn't import it on the android module. I will place it as Android-only for now.

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListView.kt
marcellogalhardo, line 133: Also, if you prefer a different format let me know. The `MEDIUM` seems like a good one IMO.

yschimke: Shouldn't we converge on dev.johnoreilly.confetti.utils.DateService? Or move away from that?

Maybe add some new simple methods there?

androidApp/src/main/java/dev/johnoreilly/confetti/utils/LocalDateTime.kt
joreilly: @marcellogalhardo I'll merge this and follow up then and see if/where date stuff might need to move to

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListView.kt
marcellogalhardo: I had a look into `DataService` and it seems we can cover all its use cases with a few extension functions inside common targeting `kotlinx.datetime` and providing an implementation per platform.

I will fill a bug for follow-up work.

androidApp/src/main/java/dev/johnoreilly/confetti/utils/LocalDateTime.kt
marcellogalhardo: Sounds good to me. Please see [here](https://github.com/joreilly/Confetti/pull/359#discussion_r1141455461).

PR #367, opened 2023-03-20
fix conference name + update play store versions

joreilly:


yschimke (approved):

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SessionsViewModel.kt
```diff
@@ -103,7 +104,7 @@ open class SessionsViewModel : KMMViewModel(), KoinComponent {
         return SessionsUiState.Success(
             conference,
             dateService.now(),
-            conference,
+            conferenceName,
```
martinbonnin, line 107: Thanks!

PR #370, opened 2023-03-20
Read conference from Mobile

yschimke:
Changes Wear to read conference from

1) Phone Settings
2) Local setting

This might be annoying if your phone isn't near by.  So I'll tweak a bit, with a clear override. Last update wins.


shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/ConferencesViewModel.kt
```diff
@@ -62,7 +64,8 @@ open class ConferencesViewModel : KMMViewModel(), KoinComponent {
         }
     }
 
-    suspend fun setConference(conference: String) {
+    open suspend fun setConference(conference: String) {
```
martinbonnin, line 67: Curious why this needs to be open. Does Koin or something else extend the ViewModel ?

androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
```diff
@@ -27,7 +28,9 @@ import org.koin.dsl.module
 val appModule = module {
     viewModelOf(::SessionsViewModel)
     viewModelOf(::AppViewModel)
-    viewModelOf(::ConferencesViewModel)
+    viewModel {
+        ConferencesViewModel { get<WearSettingsSync>().setConference(it) }
```
martinbonnin, line 32: Could `WearSettingsSync` listen to `ConfettiRepository.conferenceFlow` instead? Use that as single source of truth everywhere?

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/ConferencesViewModel.kt
joreilly: Not sure specifically about this case but KMMViewModel (https://github.com/rickclephas/KMM-ViewModel) does support view model being subcassed for example in Swift code but not sure if it's a general requirement

androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
martinbonnin: >  Should I be making this a conference listener on ConfettiRepository?

There's already `ConfettiRepository.conferenceFlow` that you can reuse. It's the one we use for background sync, etc...

> AppSettings?

Good question, I like `ConfettiRepository` better because it doesn't leak settings. But no strong preference.

martinbonnin: 👍 I think I wouldn't mind a global application-level coroutine scope that handles the wear <-> phone sync. At some point if it has app lifecycle, it's OK to have a coroutine running?

yschimke: I've moved it down to the Repository, and enabled via DI wiring.  I can't shake the feeling that a direct update on each change, is better than subscribing to a Flow.

PR #383, opened 2023-03-21
Create CONTRIBUTING.md

martinbonnin:


CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
...
+
+Contributions are very welcome.
+
+If you have a feature you want to implement or found a bug, feel free to [open an issue](https://github.com/joreilly/Confetti/issues/new) and we can discuss how to best implement/fix it.
+
+If you don't know where to start, you can also take a look at the list of [good first issues](https://github.com/joreilly/Confetti/labels/good%20first%20issue) for good places to get started with the project.
+
+Documentation, UX, design, data contributions are very welcome. 
+
+If you see a conference missing in the list, please let us know by [opening an issue](https://github.com/joreilly/Confetti/labels/good%20first%20issue).
```
oas004, line 14: Might be wrong here, but should the link here be the same as the above one (https://github.com/joreilly/Confetti/issues/new) since it is to open a new issue? :)

oas004 (approved):
Very nice! 🤩 🙌

CONTRIBUTING.md
martinbonnin: Good catch, thanks!

PR #385, opened 2023-03-21
Make AccountIcon composable less stateful

### LINKED ISSUE #375: Move state outside of AccountIcon
Currently, `AccountIcon` is very stateful and hard to preview/mock. Move that state upper in the composable tree

oas004:
Firstly, thanks for a super cool and inspiring project! 👏  A lot of cool stuff to learn for me here! :) 

In this PR I have moved the ViewModel one level up from the `AccountIcon` in an effort to try to make the `AccountIcon` composable more stateless so that it is easier to make preview functions for it.

I was not 100% sure which approach was the best for this.  I was first thinking to just pass the UIState object as a parameter. Then I thought that it might be even easier to mock if I just pass the boolean types and the user object. 

I also added a preview for light and dark theme with a background so that it was a bit easier to see.

resolves #375

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
@@ -46,9 +46,7 @@ fun AccountIcon(
         onShowSettings = onShowSettings,
         installOnWear = viewModel::installOnWear,
         updateWearTheme = viewModel::updateWearTheme,
-        user = accountUiState.user,
-        showInstallOnWear = accountUiState.showInstallOnWear,
-        isInstalledOnWear = accountUiState.isInstalledOnWear
+        uiState = accountUiState
     )
 }
 
@@ -57,14 +55,13 @@ private fun AccountIcon(
     onSwitchConference: () -> Unit,
     onSignIn: () -> Unit,
     onSignOut: () -> Unit,
+    onShowSettings: () -> Unit,
     updateWearTheme: () -> Unit,
     installOnWear: () -> Unit,
-    onShowSettings: () -> Unit,
-    user: User?,
-    showInstallOnWear: Boolean,
-    isInstalledOnWear: Boolean
+    uiState: UiState,
 ) {
     var showMenu by remember { mutableStateOf(false) }
+    val user = uiState.user
 
     IconButton(onClick = { showMenu = !showMenu }) {
         when {
@@ -124,7 +121,7 @@ private fun AccountIcon(
                 onShowSettings()
             }
         )
-        if (showInstallOnWear) {
+        if (uiState.showInstallOnWear) {
             DropdownMenuItem(
                 text = { Text("Install on Wear") },
                 onClick = {
@@ -133,7 +130,7 @@ private fun AccountIcon(
                 }
             )
         }
-        if (isInstalledOnWear) {
+        if (uiState.isInstalledOnWear) {
             DropdownMenuItem(
                 text = { Text("Update Wear Theme") },
                 onClick = {
@@ -149,6 +146,7 @@ private fun AccountIcon(
 @Preview(uiMode = UI_MODE_NIGHT_NO, name = "light theme")
 @Composable
 private fun AccountIconPreview() {
+    val accountUiState = UiState()
     ConfettiTheme {
         Surface {
             AccountIcon(
@@ -157,10 +155,8 @@ private fun AccountIconPreview() {
                 onSignOut = {},
                 updateWearTheme = {},
                 installOnWear = {},
-                user = null,
-                showInstallOnWear = false,
-                isInstalledOnWear = false,
-                onShowSettings = {}
+                onShowSettings = {},
+                uiState = accountUiState
             )
         }
     }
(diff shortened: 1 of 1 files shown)


androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -74,10 +77,22 @@ fun ConfettiScaffold(
...
                     actions = {
+                        val viewModel: AccountViewModel = getViewModel()
+                        val accountUiState = viewModel.uiState.collectAsState().value
                         AccountIcon(
                             onSwitchConference = onSwitchConference,
-                            onSignIn = onSignIn,
-                            onSignOut = onSignOut,
+                            onSignIn = {
+                                onSignIn()
+                            },
```
oas004, line 86: I think I can revert this change. Sorry about that

martinbonnin (approved):
Thanks! 

Agree on individual parameters, it makes it easier to see what the composable does 👍

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -74,10 +77,20 @@ fun ConfettiScaffold(
                         containerColor = Color.Transparent
                     ),
                     actions = {
+                        val viewModel: AccountViewModel = getViewModel()
```
yschimke, line 80: What about introducing a stateful and stateless pair? Similar to https://developer.android.com/jetpack/compose/state#state-hoisting

It keeps this class simpler.

You just end up having the original stateful form of AccountIcon, but then delegate all UI structure to the stateless form.  Then test and preview that.

oas004: Aha, yeah, that sounds nice! I can try that out! That means that I can revert the changes in the scaffold component to right?😊

yschimke: Yep.

androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
```diff
@@ -84,23 +114,43 @@ fun AccountIcon(
                 onSwitchConference()
             }
         )
-        if (uiState.showInstallOnWear) {
+        if (showInstallOnWear) {
```
yschimke, line 117: Maybe a personal preference but I'd pass the uiState in and create it in the previews.

oas004: As I wrote in the description, I was not 100% sure, but what is the reason that you would have chosen that approach? :)

martinbonnin: I'm almost sure I saw a Google video advocating for individual parameters but can't find it anymore. Either works for me

yschimke: I don't think it's the case.  I'll try to find clear guidance.

A few reasons

1) It's really unwieldy, and doesn't scale well as features get more complex.  You end up changing a lot of params up and down the call chain. The Composable contract is also specific to the current UI, so it must change and break callers every time you add some additional field.
2) UiState might be simple now, but over time, you might end up with a union of types.  Loading, Failed, SignedIn, Anonymous.  If these each have different valid params, then your composable with individual fields, every param becomes nullable.
3) to me it reads cleaner

yschimke: I feel like they are generally pretty consistent about this

https://github.com/android/nowinandroid/blob/2388d4eb2417695c41fabd3fa916d899614ccd9f/feature/interests/src/main/java/com/google/samples/apps/nowinandroid/feature/interests/InterestsScreen.kt#L55

XScreen, XUiState, XRoute.

Although they do pick out bits of that UIState and pass it into the individual screen components.  But that makes sense, you only need part of data.

martinbonnin: > XScreen, XUiState, XRoute

+1 on this naming, makes a lot of sense to me. 

In that very specific `AccountIcon` case it doesn't really apply because `AccountIcon` is not a screen. Having a ViewModel there makes it impossible to preview `SessionsView` for an example (which should really have been `SessionsScreen`).

Question: should we juste remove the `AccountViewModel` altogether and move that state to the upmost "Route" level?

oas004:
☝️ Had to force push after resolving conflicts from this https://github.com/joreilly/Confetti/pull/388 :)

androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
yschimke: I think so.

yschimke:
@oas004 you shouldn't have to force push, but not a big issue.

oas004:
Okay, sorry about that. I tried to just push first, but I was rejected because "the tip of my current branch is behind" in git. Is it better to merge commit or something like that instead of rebase?

yschimke:
@oas004 for your own PR branch, it doesn't really matter.  But yep, merge commits is a simpler workflow.

oas004:
Okay, I'll keep that in mind for the future. Thanks! :)

androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
oas004: I can fix this, do you want me to do that in this PR? Or is it better to separate the removal of the ViewModel? :)

PR #388, opened 2023-03-22
Theme settings screen

joreilly:
Screenshot_20230322_085456


martinbonnin:
Nice! Can there be a dark overlay behind the dialog? Or thinking forward, should we turn the settings into a standalone screen? If we add stuff like developer settings, we'll need to make the Dialog scrollable which sounds a bit weird? Thoughts?

joreilly:
Yeah, was thinking that as well.....that we'd likely need to move to standalone screen.

PR #403, opened 2023-03-23
Add `share` button

### LINKED ISSUE #303: Share option
Talks could support a share button, so I can easily communicate with friends and send them the talks I'm interested to.

A power function could be to share N number of talks (by hold click, select multiple ones and share) or share all my bookmarks - but that can be done separately if we think it is worth to build it. These power function ideas are nice to have, but not a must.

marcellogalhardo:
Naive implementation for share option. It includes a share icon button at the top bar, and clicking on it triggers a default send text intent, which contains the basic information from a session (I'm not sure if the details description should be shared - it can be too long - but for now it is included). You can send it to any "text" receiving app (WhatsApp, e-mail, etc).

The current implementation only works with Android, and it is a simple function interacting with the Android context.

Fixes #303

| Share Button | Share Bottom-Sheet | Shared Content |
|---|---|---|
| Screenshot_1679612384 | Screenshot_1679612388 | Screenshot_1679612394 |


androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
```diff
@@ -146,7 +146,7 @@ fun Chip(name: String) {
 private fun rememberShareDetails(details: SessionDetails?): () -> Unit {
     val context = LocalContext.current
 
-    return remember(details) {
+    return remember(context, details) {
```
martinbonnin, line 149: Out of curiosity, why is `context` needed?

marcellogalhardo: I added it because I'm capturing the context within the block - but to be fair, I don't think it is required.

marcellogalhardo:
Uhhh, sorry for the formatting noise. If needed, I can fix it tomorrow. 🙃

PR #428, opened 2023-03-25
Improve Search experience

marcellogalhardo:
* ~Search is now always visible.~
* Search removes any empty dates, they are unnecessary clusters.
* Pressing back now closes the search, before closing the app.
* Account action moved to the right side of the top bar.

~@joreilly + @martinbonnin I'm not sure how much I like the new search compared with the previous one (and there is no animation now), but it will make it easier to include a filter action in the future (on the right side of the account).~

EDIT: I have dropped the new Search UI for now, we can discuss it separately.


marcellogalhardo:
I have reverted the search UX, for now - we can discuss it separately.

PR #429, opened 2023-03-25
Add Multiplatform Logging

marcellogalhardo:
While contributing to the multiplatform client code, I missed a logging solution. I'm not sure if `Napier` is the best solution out there but it is the only one I found supporting all platforms we support.

https://github.com/AAkira/Napier

What do you folks think of it?

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
@@ -1,18 +1,16 @@
 package dev.johnoreilly.confetti
 
 import android.app.Application
-import androidx.work.ExistingPeriodicWorkPolicy
-import androidx.work.WorkManager
 import coil.ImageLoader
 import coil.ImageLoaderFactory
-import com.google.firebase.FirebaseApp
 import com.google.firebase.crashlytics.ktx.crashlytics
 import com.google.firebase.crashlytics.ktx.setCustomKeys
 import com.google.firebase.ktx.Firebase
-import dev.johnoreilly.confetti.di.initKoin
 import dev.johnoreilly.confetti.di.appModule
-import dev.johnoreilly.confetti.work.RefreshWorker
+import dev.johnoreilly.confetti.di.initKoin
 import dev.johnoreilly.confetti.work.setupDailyRefresh
+import io.github.aakira.napier.DebugAntilog
+import io.github.aakira.napier.Napier
 import org.koin.android.ext.android.get
 import org.koin.android.ext.koin.androidContext
 import org.koin.android.ext.koin.androidLogger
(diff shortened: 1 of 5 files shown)


joreilly:
No strong opinion....I'm using Kermit on some other projects but Napier also popular as well.

martinbonnin:
Apollo doesn't do logging but exposes APIs instead (see also https://publicobject.com/2022/05/01/eventlisteners-are-good/). If I need to debug something quickly, I usually do `println()` and (hopefully!) remove it before committing.

But Napier works too 👍

martinbonnin (approved):

androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
```diff
@@ -32,7 +30,10 @@ class ConfettiApplication : Application(), ImageLoaderFactory {
         } else {
             Firebase.crashlytics.setCrashlyticsCollectionEnabled(false)
         }
-        
+
+        // Initialize Logging.
+        Napier.base(DebugAntilog())
```
joreilly, line 35: Does something equivalent have to be done for Wear and iOS clients/

marcellogalhardo: For iOS:
```
fun debugBuild() {
    Napier.base(DebugAntilog())
}
```

We would also need to do it on WearOS + Compose Desktop.

marcellogalhardo:
@martinbonnin I'm familiar with the article, great one btw.

Does `print` in a multiplatform code (i.e., SessionsViewModel) print on Logcat console? It wasn't showing for me (but I don't have much experience with KMP anyway, so I may have been doing it wrong?). 🤔

androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
marcellogalhardo: I will include it on Wear and Desktop, I don't have iOS tooling to compile and test, tho.

compose-desktop/src/main/kotlin/main.kt
```diff
@@ -36,6 +54,11 @@ private val dateService = JvmDateService()
 fun main() = application {
     val windowState = rememberWindowState()
 
+    LaunchedEffect(key1 = this) {
```
marcellogalhardo, line 57: @joreilly I have no clue if that is the "correct way" to do it on Compose Desktop. Suggestions?

androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
marcellogalhardo: Included on Android, Wear, Desktop. I currently don't have the tooling to compile and test iOS, someone will need to do it there. 😄

martinbonnin:
> Does print in a multiplatform code (i.e., SessionsViewModel) print on Logcat console?

It should although it bypasses the logcat tags (and maybe other things?) so it will end up with a `System.out` tag. If you're filtering on `System.out` you should see it (which is also a good way to filter out the noise :) )

compose-desktop/src/main/kotlin/main.kt
joreilly: I'm definitely no expert on Compose for Desktop but that seems ok to me at least

PR #433, opened 2023-03-25
Screenshot test for Speaker Details Screen

yschimke:


yschimke:
There is some timing issue with the scrollAway modifier in Robolectric, or maybe it's always there at app startup even on devices. But instead of solving, just make it configurable for now.

PR #434, opened 2023-03-25
Tweak gradle-build-action usages

Goooler:


yschimke:
What's the context behind this change? Is it faster? If the reverse PR was put up, I'd be inclined to accept so some context would help.

Goooler:
I see there are 3 `gradle-build-action`s used in `build-android` job, only 1 is enough in 1 job.

martinbonnin (approved):

PR #436, opened 2023-03-25
Switch to Firebase Auth for Wear.

yschimke:


yschimke:
Got a failure, so holding off

`com.google.firebase.auth.FirebaseAuthInvalidCredentialsException: The supplied auth credential is malformed or has expired.`

PR #438, opened 2023-03-26
Add a Conference A11y test

yschimke:
Adds A11y overlay and legend to the existing Conference screen.

image

Not clear we want to do this, but it follows on from the approaches in

https://github.com/google/horologist/blob/main/paparazzi/src/main/java/com/google/android/horologist/paparazzi/a11y/A11ySnapshotHandler.kt
and 
https://github.com/cashapp/paparazzi/pull/555


yschimke:
Most of the PR is reusable framework code.

The actual conference test https://github.com/joreilly/Confetti/pull/438/files#diff-b6dab13f09410aa4f6acdc2e9cf5a84ac7a5b0eb1c44d15d9b0220e5d60b6139 shows how easy it is to add an A11y test to a single screen.

For tall screens, we may want to do something similar to

https://github.com/google/horologist/blob/main/media-ui/src/test/snapshots/images/com.google.android.horologist.media.ui.screens.browse_PlaylistDownloadBrowseScreenA11yTallScreenshotTest_browseScreen.png

martinbonnin:
> Not clear we want to do this

What's the drawback? Adding complexity to the codebase? On the plus side, it's a very nice incentive to get a11y working when adding screens.

yschimke:
Let's do it then.

martinbonnin (approved):

oas004:
This is really cool! So inspiring! 🤩

Issue #391, opened 2023-03-22
Remove the AccountViewModel and move that state to the upmost "Route" level


oas004:
Making this issue since we discussed it in the thread here (https://github.com/joreilly/Confetti/pull/385#discussion_r1144735106) :) 
I could not find it in the issues. Hope that is okay :)

oas004:
I was wondering a bit about this task. If we remove the AccountViewModel, should then the `SessionsViewModel` expose more than one uiState? Or should it be part of the successState that of the sessions? If that is the case should the accountIcon not be shown if there is an error loading the sessions? :)

Is there something here that I'm missing or any other way that is prefered? @martinbonnin

yschimke:
Yeah, I think a rule that screens have a 1:1 viewmodel makes sense. And like nowinandroid, if uiState updates separately and is used for different parts then it makes sense as 2 flows

oas004:
Okay that makes sense, I'll do that approach then😊

martinbonnin:
Agree with 1 viewmodel per screen. No strong preference whether we want 2 different states or just 1. I guess if the state maps to nested composables, it's easier to make 2, either works for me. 

If we go with 2 different states, I would rename `uiState` to have `sessionsState` & `accountState` (or something like this).

oas004:
I ran into another issue with trying out this. When migrating the functionallity from AccountViewModel to SessionViewModel, the AccountViewModel is injecting Authentication class and WearSettingSync class. These are in the `:androidApp` module but the SessionViewModel is in the shared mobile module. This means I have to move them into that module to avoid a circular dependency issue for the ViewModel if I'm not mistaken. These classes are using some dependencies that I don't think I can implement in the common module.

Was there a reason that the AccountViewModel was not shared between Android and iOS? Is the login part handled differently on iOS? :)

oas004:
I guess I can extend SessionsViewModel and add the Android behavior to that, is that okay? What do you think about that? :)

yschimke:
@oas004 I'd extend as you suggest.

But honestly you are in uncharted territory, so this will be a learning experience for what works well.

oas004:
I got something working, but I'm not 100% sure about the approach. Made a draft in #446

oas004:
Can I close this issue as completed? Or was there anything left regarding this? :) I opened an issue regarding Robolectric for Android if we are going to add integration tests for this :) #522

martinbonnin:
All good, thanks for diving into this!

PR #446, opened 2023-03-26
Remove account VM and move state to top-level VM

oas004:
This should solve #391 

I'm not 100% sure on this approach, since the AccountViewModel is using Android specific stuff I could not move this to the mobile shared module. So here I'm extending the VMs in the Android module and adding that stuff and making that the VM that the view is using.

I'm not that sure about this approach, maybe I can move the behaviour to it's own class since here I feel that I'm duplicating a lot of stuff. Any opinions on this is much appreciated.

Might be that having a separate VM for account is something that we should have 🤔


androidApp/src/main/java/dev/johnoreilly/confetti/search/SearchAndroidViewModel.kt
```diff
@@ -0,0 +1,51 @@
...
+                wearSettingsSync.installOnWearNode(it.id)
+            }
+        }
+    }
+
+    fun updateWearTheme() {
+        viewModelScope.coroutineScope.launch {
+            wearSettingsSync.updateWearTheme()
+        }
+    }
```
oas004, line 50: Maybe I can extract this somehow?

androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
```diff
@@ -33,9 +35,11 @@ val appModule = module {
     viewModelOf(::SpeakersViewModel)
     viewModelOf(::SessionDetailsViewModel)
     viewModelOf(::SpeakerDetailsViewModel)
-    viewModelOf(::AccountViewModel)
     viewModelOf(::SettingsViewModel)
     viewModelOf(::SearchViewModel)
+    viewModelOf(::SessionAndroidViewModel)
+    viewModelOf(::SpeakerAndroidViewModel)
+    viewModelOf(::SearchAndroidViewModel)
```
oas004, line 42: This works runtime, but it gives me an red line in the IDE since it does not directly extend the KMMViewModel abstract class

yschimke: I got this too. I haven't raised a bug, but unless there is a simple fix, maybe we should.

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerAndroidViewModel.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    fun installOnWear() {
+        viewModelScope.coroutineScope.launch {
+            accountUiState.value.wearNodes.filter { !it.isAppInstalled }.forEach {
+                wearSettingsSync.installOnWearNode(it.id)
+            }
+        }
+    }
+
+    fun updateWearTheme() {
```
yschimke, line 40: I think this duplication is a step in the wrong direction.  I wonder if we can split this into Scaffold and Individual screens.  But as is, I wouldn't add this.  It feels like we have got the lines drawn in slightly the wrong place currently, and it's visible in your PR.

How should screens have common functionality for android?

martinbonnin: > I wonder if we can split this into Scaffold and Individual screens

Some screens ("login", "conferences") do not have a `Scaffold` because they use the whole screen. But even if we had, the 1:1 mapping between screen and viewmodel somewhat enforces duplication of `FooViewModel.updateWearTheme()` and others...

Maybe for account/wear we use singleton (maybe even POJO?) state holders 

```kotlin
    val viewModel: SessionAndroidViewModel = getViewModel<SessionAndroidViewModel>()
    val uiState by viewModel.uiState.collectAsStateWithLifecycle()

    // account and wear we can reuse in every screen that has an account button
    val accountHolder: AccountHolder = koinInject()
    val accountUiState by accountHolder.state.collectAsStateWithLifecycle()

    val wearHolder: WearHolder = koinInject()
    val wearUiState by wearHolder.state.collectAsStateWithLifecycle()

```

martinbonnin:
> the AccountViewModel is using Android specific stuff I could not move this to the mobile shared module.

I was thinking we could either:

1. Use [firebase-multiplatform](https://firebaseopensource.com/projects/gitliveapp/firebase-kotlin-sdk/) 
Or
2. do `expect`/`actual` 

to keep the ViewModel 100% in shared code. 2. sounds certainly less risky than 1. but not sure. I'll dig a bit tomorrow, see what's possible.

martinbonnin:
PR to use Firebase MPP there: https://github.com/joreilly/Confetti/pull/449.

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerAndroidViewModel.kt
oas004: I can try it out with the stateHolder, that would at least reduce the duplication. If I'm splitting up the Scaffold with it's own VM I think I'm kinda back to AccountViewModel.

oas004:
Super cool! 🥳 Do you think it could be possible to also move the WearSettingsSync to common as well? Or should I try to introduce the stateholder strategy for that? :)

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerAndroidViewModel.kt
martinbonnin: I'm still unclear whether Account/Wear must be ViewModels in the `KMMViewModel`. The advantage is that we get the nice `@UiState` for iOS out of the box but it feels like we're abusing a bit the ViewModel cause these are things that are not specifically bound to a View.

oas004: Very nice PR! Really cool to see how you solved it! :)

martinbonnin: Well TBH I'm not sure. This solves the bookmark sync issue but it makes everything more verbose because we now need to pass `uid` (and tokenProvider) all the way down. Let's wait for @yschimke input and if he's ok, let's do something similar for `Wear`? Maybe even using `WearSettingsSync` and `koinInject` directly from the `ConfettiScafflod` ?

oas004: Yeah let's do that :) Is there any guidelines from koin to use the `koinInject`  directly in Compose like this? I could not find anything on their pages. Do you know? :)

oas004: I think I found some here: https://insert-koin.io/docs/reference/koin-compose/compose although they where not that describing, I think you are correct :)

oas004:
Should I maybe rebase this PR on #453 and do the `koinInject` approach for WearSettings? Or do you prefer that I close this PR? :)

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerAndroidViewModel.kt
yschimke: Sorry, missed this.

yschimke: I'm very open to the koinInject. There is an open debate in the compose world on the value of ViewModels, especially when Compose is used cross platform.  Maybe time to explore this ourselves.

martinbonnin:
> Should I maybe rebase this PR on https://github.com/joreilly/Confetti/pull/453 and do the koinInject approach for WearSettings? 

Yep, let's do this 👍  I merged #453 to avoid having too many dependent PRs in flights.

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerAndroidViewModel.kt
martinbonnin: I think part of the question for Confetti is how to bridge that to SwiftUI as we're reluying on  https://github.com/rickclephas/KMM-ViewModel to bridge to observable SwiftUI models. But since Wear is most likely Android only we don't need KMMViewModels there.

androidApp/src/main/java/dev/johnoreilly/confetti/speakers/SpeakerView.kt
```diff
@@ -60,6 +54,10 @@ fun SpeakersRoute(
         onSignIn = onSignIn,
         onSignOut = onSignOut,
     ) {
+        val viewModel: SpeakersViewModel = getViewModel<SpeakersViewModel>().apply {
```
oas004, line 57: Maybe I dont need this change 🤔

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -60,10 +64,32 @@ fun ConfettiScaffold(
...
+                wearSettingSync.installOnWearNode(it.id)
+            }
+        }
+    }
+
+    fun updateWearTheme() {
+        coroutineScope.launch {
+            wearSettingSync.updateWearTheme()
+        }
+    }
```
oas004, line 91: This I'm not 100% sure about 🤔

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SearchViewModel.kt
```diff
@@ -12,7 +12,7 @@ import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 
-class SearchViewModel(
+open class SearchViewModel(
```
oas004, line 15: This I need to revert

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
martinbonnin: My hunch is that this should run in some global scope, not the composable scope. If you move out of the composable, you probably don't want to cancel the installation. I'd go with something like this directly inside `WearSettingsSync` but @yschimke will know better.

```kotlin
    private var job: Job? = null

    // public non-suspend function to be called from button callbacks, etc...
    fun installOnWearNode(nodeId: String) {
        if (job != null) {
            // already installing, skip
            return
        }
        
        // coroutineScope at this point is a app-global scope
        job = coroutineScope.launch {
            installOnWearNodeInternal(nodeId)
            
            job = null
        }
        
    }
    private suspend fun installOnWearNodeInternal(nodeId: String) { ... }
```

yschimke: I guess three options.

a) the composable is responsible for this completing. So we should model the inprogress state even a spinner.
b) since it's going to be almost immediate, we actually block on it completing but don't show anything.
c) We move it up to a wider scope.  We just lost view model scope, so maybe the global scope for now?

oas004: hm, maybe I can provide the activity lifecycle scope and use that? I feel like using globalScope is a bit heavy and according to the documentation, easy to accidentally create resource or memory leaks :)

martinbonnin: I liking providing a small UI or blocking: gives feedback to the user that something is happening but this is more involved and can be done in a separate PR. 

Re: global scope, I had [this one](https://github.com/joreilly/Confetti/blob/ba7b222b5bcc211a15b4bad7ab8ce1a92b80d6df/androidApp/src/main/java/dev/johnoreilly/confetti/wear/WearSettingsSync.kt#L29) in mind but activityScope works too!

oas004: Aha sorry I misunderstood that. I think that can be used :)

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -60,10 +64,30 @@ fun ConfettiScaffold(
...
+    fun installOnWear() {
+        wearSettingNodeState.filter { !it.isAppInstalled }.forEach {
+            wearSettingSync.installOnWearNode(it.id)
+        }
+    }
+
+    fun updateWearTheme() {
+        coroutineScope.launch {
+            wearSettingSync.updateWearTheme()
+        }
```
martinbonnin, line 88: Can you move this to `WearSettingsSync`, similar to `installOnWear`? I think the logic and scoping should be pretty similar

oas004: Aha! Yes good idea!

martinbonnin (approved):
LGTM! Would be nice to integration test this but I guess it's not easy :)

oas004:
Yeah I agree! That would be nice to add! Maybe we can add an issue on that as well as the loading indication to the user? :)

martinbonnin:
Yup, issues are always good !

oas004:
Are we btw running android integration tests on CI for this project?  Is that something we want to setup? :)

yschimke:
> Are we btw running android integration tests on CI for this project?  Is that something we want to setup? :) 

They have been too flaky for me in GitHub CI, especially once complex enough to capture these types of code paths.

Instead I'm using robolectric app tests on wear. There is some koin/robolectric cleanup issue. But generally works well including sqllite and OkHttp. But we should mock that out

martinbonnin:
I tried to manually integration test this but no matter what I do, I get `wearNodes.find { it.isAppInstalled } == null`, i.e. installing works well but the app is not detected on the watch after that so I can't test updating the theme. @yschimke , do I need to run the same keystore for both mobile & wear OS app or something like this? (Currently running a debug mobile app and installing the release app on the watch)

yschimke:
@martinbonnin I'm changing that at the moment, want to leave it to me?

If debug and release are using different signing keys, that will never work.

But also if you are using Android Studio canary, the pairing flow is broken until the next canary, 12?

To confirm this, on the watch go to settings, google accounts and try to add a new account.  If it can't connect, that's the issue.

martinbonnin:
> I'm changing that at the moment, want to leave it to me?

Sure, go for it!

> But also if you are using Android Studio canary, the pairing flow is broken until the next canary, 12?

I was using a real watch this time, let me check with the emulator

martinbonnin:
Alright, I got this with the emulator. 

Screenshot 2023-04-02 at 16 46 02

yschimke:
If you have emulator binary 33.1.3, then it's a known issue.

You can fix, but it's involves a second stable copy or Android Studio, uninstalling, reinstalling the emulator binary, and then running the emulators from a terminal instead of Android Studio.

Or wait for the fix.

martinbonnin:
I had a much older version of the emulator (32.1.7). But no rush, I can use hardware device for the time being. I'll test with 2 debug apps

yschimke:
That should work.

martinbonnin:
I have my watch correctly detected now but for some reason the theme does not update when clicking. I'm going through `settingsDataStore.updateData` but nothing changes on the watch. I shouldn't need to restart the watch app, right?

joreilly:
>I have my watch correctly detected now but for some reason the theme does not update when clicking. 

Not sure if it's the same thing but I had to have "Dynamic Color" set to Yes in settings for that to work.

martinbonnin:
> Not sure if it's the same thing but I had to have "Dynamic Color" set to Yes in settings for that to work.

Right on point 🎯 👍 . Changing the wallpaper changes the watch theme.

yschimke:
@oas004 sorry for the conflicts, didn't think about it.

oas004:
No worries! 😊 I'll try to resolve conflicts when I get home tonight 😊

oas004:
@yschimke I hope I managed to solve the conflicts correctly now. Could you take a look when you have time? :)

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -61,10 +65,22 @@ fun ConfettiScaffold(
     content: @Composable (ScaffoldState) -> Unit,
 ) {
     val authentication = koinInject<Authentication>()
+    val wearSettingSync = koinInject<WearSettingsSync>()
```
yschimke, line 68: Don't we create multiple of these, seem to be included in 4 different screens.  Would it be worth hoisting this to the top level VM?

```
fun ConfettiApp(
    navController: NavHostController,
    windowSizeClass: WindowSizeClass,
    displayFeatures: List<DisplayFeature>,
) {
    val appState: ConfettiAppState = rememberConfettiAppState(
        windowSizeClass,
        displayFeatures,
        navController
    )
```

So we only do this once for activity?

androidApp/src/main/java/dev/johnoreilly/confetti/wear/WearSettingsSync.kt
```diff
@@ -115,7 +118,21 @@ class WearSettingsSync(
         }
     }
 
-    suspend fun installOnWearNode(nodeId: String) {
+    // public non-suspend function to be called from button callbacks, etc...
+    fun installOnWearNode(nodeId: String) {
```
yschimke, line 122: This is pretty quick, I wouldn't necessarily worry about the Job. Running it twice is unlikely and probably safe.

yschimke: Probably a style thing, but what about this just being a suspend function, and using a higher level scope when calling it from a VM?

yschimke (approved):
Nothing blocking, just some stylistic questions.

androidApp/src/main/java/dev/johnoreilly/confetti/wear/WearSettingsSync.kt
oas004: But can I then use the `coroutineScope` when calling it from compose? Like

```
    val scope = rememberCoroutineScope()
    fun installOnWear() {
        wearSettingNodeState.filter { !it.isAppInstalled }.forEach {
            scope.launch {
                wearSettingSync.installOnWearNode(it.id)
            }
        }
    }

```

Because I think that was what we were a bit worried about that that scope would be easily cancelled and canceling the call 🤔

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
oas004: I think that could be a good idea! Am I understanding this correct that it should be added as part of the AppState then right? Not passed through the composables? :)

androidApp/src/main/java/dev/johnoreilly/confetti/wear/WearSettingsSync.kt
yschimke: I was suggesting pulling it up to the app ViewModel.

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
yschimke: For me, yep.

martinbonnin: What about registering it as a singleton in DI? If we know we want only one, we could enforce it there instead of having an `AppState` that grows forever? Then we have Koin modules growing forever but we need to register the constructor there anyway, might as well make it `single {}`?

martinbonnin: Actually looks like [it's already the case](https://github.com/joreilly/Confetti/blob/afecf03f840941d0f04962c3a099084c755876fc/androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt#L59)?

androidApp/src/main/java/dev/johnoreilly/confetti/wear/WearSettingsSync.kt
martinbonnin: FWIW, `AppViewModel` is really `InitialLoadingViewModel` in the model where there is a 1:1 relationship between VM and screens. I plead guilty of very bad naming there. 

If we introduce an app-level VM, it'd be parallel to `AppState` I guess and used from the ConfettiApp composable.
That would break the 1:1 screen:VM relationship I was operating under but that'd work for me too, we can be pretty flexible if we don't start creating a VM for each single composable.

yschimke: Probably better we head for a common understanding and approach. Feels free to ship this and dismiss my suggestions. Nothing blocking.

martinbonnin: 👍 Merging this PR, let's follow up in separate PRs for style/aligment.

PR #449, opened 2023-03-27
Use Firebase mpp SDK and move Authentication to shared

martinbonnin:
Follow up PR is to move `Authentication` outside of the VM so that it can be shared more easily between screens


joreilly:
Does this potentially make it easier to use from iOS client?

martinbonnin:
@joreilly in the short term, looks like it broke iOS :( 

```
The /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld command returned non-zero exit code: 1.
output:
ld: framework not found FirebaseAuth
```

In the long term I'm hoping we can share some of the auth logic indeed. I'll revisit shortly

androidApp/src/main/java/dev/johnoreilly/confetti/account/SignInView.kt
```diff
@@ -16,12 +16,15 @@ import androidx.compose.ui.platform.LocalContext
...
+import org.koin.compose.koinInject
 
 
 @Composable
 fun SignInRoute(onBackClick: () -> Unit) {
     Box(
-        modifier = Modifier.fillMaxWidth().fillMaxHeight()
+        modifier = Modifier
+            .fillMaxWidth()
+            .fillMaxHeight()
```
yschimke, line 27:
```suggestion
            .fillMaxSize()
```

yschimke (approved):

androidApp/src/main/java/dev/johnoreilly/confetti/account/SignInView.kt
martinbonnin: TIL, thanks!

PR #453, opened 2023-03-27
Auth rework

### LINKED ISSUE #450: Signing in does not refresh bookmarks

martinbonnin:
Follow up from https://github.com/joreilly/Confetti/pull/449

This turns `User` and `Authentication` into interfaces that we can fake for tests/previews. Also fixes #450

@oas004 @yschimke let me know what you think, it passes `uid` and `tokenProvider` all the way down to the repository so now we have an `uid` parameter everywhere in addition to `conference` (but not in deeplinks). It's a bit verbose (Compose loves long lists of arguments) but at least it's correct and the bookmark state is now in-sync with the AccountIcon.

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
@@ -24,6 +24,7 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import coil.compose.AsyncImage
+import dev.johnoreilly.confetti.auth.User
 import dev.johnoreilly.confetti.ui.ConfettiTheme
 import org.koin.androidx.compose.getViewModel
 
@@ -33,20 +34,19 @@ fun AccountIcon(
     onSignIn: () -> Unit,
     onSignOut: () -> Unit,
     onShowSettings: () -> Unit,
-    viewModel: AccountViewModel = getViewModel()
+    user: User?,
+    viewModel: AccountViewModel = getViewModel(),
 ) {
     val accountUiState = viewModel.uiState.collectAsState().value
     AccountIcon(
         onSwitchConference = onSwitchConference,
         onSignIn = onSignIn,
-        onSignOut = {
-            viewModel.signOut()
-            onSignOut()
-        },
+        onSignOut = onSignOut,
         onShowSettings = onShowSettings,
         installOnWear = viewModel::installOnWear,
         updateWearTheme = viewModel::updateWearTheme,
-        uiState = accountUiState
+        uiState = accountUiState,
+        user = user,
     )
 }
 
@@ -59,9 +59,9 @@ private fun AccountIcon(
     updateWearTheme: () -> Unit,
     installOnWear: () -> Unit,
     uiState: UiState,
+    user: User?,
 ) {
     var showMenu by remember { mutableStateOf(false) }
-    val user = uiState.user
 
     IconButton(onClick = { showMenu = !showMenu }) {
         when {
@@ -153,10 +153,22 @@ private fun AccountIconPreview() {
                 onSwitchConference = {},
                 onSignIn = {},
                 onSignOut = {},
+                onShowSettings = {},
                 updateWearTheme = {},
                 installOnWear = {},
-                onShowSettings = {},
-                uiState = accountUiState
+                uiState = accountUiState,
+                user = object: User {
+                    override val name: String
+                        get() = "Foo"
+                    override val email: String?
+                        get() = "foo@bar.com"
+                    override val photoUrl: String?
+                        get() = null
+                    override val uid: String
+                        get() = "123"
+
+                    override suspend fun idToken(forceRefresh: Boolean): String? = null
+                }
             )
         }
     }
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
@@ -167,7 +167,7 @@ private fun AccountIconPreview() {
                     override val uid: String
                         get() = "123"
 
-                    override suspend fun idToken(forceRefresh: Boolean): String? = null
+                    override suspend fun token(forceRefresh: Boolean): String? = null
                 }
             )
         }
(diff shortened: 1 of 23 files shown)


oas004:
I think this makes sense! Now you are scoping the User specifically to the scaffold? Is it always the case that if you are using ConfettiScaffold you have to implement the AccountIcon in the toolbar? :)

androidApp/src/main/java/dev/johnoreilly/confetti/search/SearchRoute.kt
```diff
@@ -21,8 +24,9 @@ fun SearchRoute(
     onSignOut: () -> Unit,
 ) {
     val viewModel = getViewModel<SearchViewModel>()
+    val user by koinInject<Authentication>().currentUser.collectAsStateWithLifecycle()
     SideEffect {
-        viewModel.configure(conference)
+        viewModel.configure(conference, user?.uid, user)
```
oas004, line 29: Just wondering, Is there a case where the uid is different from the user.uid? :)

martinbonnin: Never 😅 . It's just that `User` is not available in `commonMain`, only in `mobileMain` so we can't pass it all the way down. I guess it also prevents leakage of `photoUrl`, `name` and `email` which aren't really needed but we could certainly put everything in only one parameter at some point.

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SessionsViewModel.kt
```diff
@@ -128,28 +78,38 @@ open class SessionsViewModel : KMMViewModel(), KoinComponent {
...
+        job?.cancel()
+        job = viewModelScope.coroutineScope.launch {
+            responseData(showLoading, forceRefresh).collect {
+                responseDatas.send(it)
+            }
+        }
     }
 
-    suspend fun refresh() = refresh(false)
+    suspend fun refresh() = refresh(false, true)
```
oas004, line 112: This was a bit hard to follow to be honest, maybe it would be nice with a named argument IMO at least here :)

androidApp/src/main/java/dev/johnoreilly/confetti/search/SearchRoute.kt
oas004: Aha okay, yeah I was just a bit confused since you pass both here :)

martinbonnin:
> Now you are scoping the User specifically to the scaffold? Is it always the case that if you are using ConfettiScaffold you have to implement the AccountIcon in the toolbar?

`ConfettiScaffold` has the `AccountIcon` as a child Composable always so if you have a `ConfettiScaffold`, you have an `AccountIcon`. And therefore the screen must handle goto Signin/Conferences/Signout/etc...

shared/src/jvmMain/kotlin/dev/johnoreilly/confetti/di/KoinJVM.kt
```diff
@@ -29,4 +29,4 @@ actual fun platformModule() = module {
     }
 }
 
-actual fun getDatabaseName(conference: String) = "jdbc:sqlite:$conference.db"
+actual fun getDatabaseName(conference: String, uid: String?) = "jdbc:sqlite:$conference$uid.db"
```
yschimke, line 32: This is surprising, do we need multiple concurrent users?

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SearchViewModel.kt
```diff
@@ -77,8 +77,8 @@ class SearchViewModel(
             }
     }
 
-    fun configure(conference: String) {
-        sessionsViewModel.configure(conference)
+    fun configure(conference: String, uid: String?, tokenProvider: TokenProvider?) {
```
yschimke, line 80: Is this a limitation of shared view models? We need an explicit configure?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ApolloClientCache.kt
```diff
@@ -19,64 +27,71 @@ interface TokenProvider {
...
         }
     }
 
     private fun clientFor(
         conference: String,
+        uid: String?,
         writeToCacheAsynchronously: Boolean
     ): ApolloClient {
-        val sqlNormalizedCacheFactory = SqlNormalizedCacheFactory(getDatabaseName(conference))
+        val sqlNormalizedCacheFactory = SqlNormalizedCacheFactory(getDatabaseName(conference, uid))
```
yschimke, line 81: Is this basically to avoid accidental contamination across users?  Could we do something more extreme, like restart the app (if possible) and delete the DB, to avoid maintaining per user files?

androidApp/src/main/java/dev/johnoreilly/confetti/account/AccountIcon.kt
```diff
@@ -153,10 +153,22 @@ private fun AccountIconPreview() {
                 onSwitchConference = {},
                 onSignIn = {},
                 onSignOut = {},
+                onShowSettings = {},
                 updateWearTheme = {},
                 installOnWear = {},
-                onShowSettings = {},
-                uiState = accountUiState
+                uiState = accountUiState,
+                user = object: User {
```
yschimke, line 160: Maybe add a FakeUser type in src/debug?

yschimke (approved):
Overall I like passing the user down through from the Scaffold.  Makes sense to load it once.

Not hugely happy about the per user files, but may be required. I can image big production apps have this sort of complication to avoid mixing the streams.

A small nit, you don't have deeplinks for android, just routes.  Deeplinks would let you open a url.  See an example here https://github.com/joreilly/Confetti/blob/main/wearApp/src/main/java/dev/johnoreilly/confetti/wear/sessiondetails/navigation/SessionDetailsDestination.kt#L54

shared/src/jvmMain/kotlin/dev/johnoreilly/confetti/di/KoinJVM.kt
martinbonnin: It's more to avoid cloberring the bookmarks cache between users. If we don't do this, we have to keep track of the uid and remove everything on logout to make sure we don't load wrong data if someone else logs in.

martinbonnin:
Merging this for now for @oas004 to rebase on top of it and avoid potential conflicts. Will follow up with:

* Only one DB for users
* Fake user for debug

PR #459, opened 2023-03-28
Add pagination to the speakers field (backend only)

BoD:
Adds pagination to speakers on the backend. This is not really useful for the mobile app at the moment (where I've passed `first: 1000`) but could be useful in the future. Feel free to ignore it if it's more trouble than usefulness 😅

### DIFF
--- a/backend/service-graphql/src/main/kotlin/dev/johnoreilly/confetti/backend/graphql/DataStoreDataSource.kt
+++ b/backend/service-graphql/src/main/kotlin/dev/johnoreilly/confetti/backend/graphql/DataStoreDataSource.kt
@@ -22,6 +22,7 @@ fun DConfig.toConference(): Conference {
         days = days,
     )
 }
+
 class DataStoreDataSource(private val conf: String, private val uid: String? = null) : DataSource {
     private val datastore = DataStore()
 
@@ -55,7 +56,7 @@ class DataStoreDataSource(private val conf: String, private val uid: String? = n
     }
 
     private val _speakers by lazy {
-        datastore.readSpeakers(conf).map { it.toSpeaker() }
+        datastore.readSpeakers(conf).map { it.toSpeaker() }.sortedBy { it.name }
     }
 
     private val sessionCache = mutableMapOf<String, Session>()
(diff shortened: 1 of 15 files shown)


backend/service-graphql/src/main/kotlin/dev/johnoreilly/confetti/backend/graphql/DataStoreDataSource.kt
```diff
@@ -44,9 +54,6 @@ class DataStoreDataSource(private val conf: String, private val uid: String? = n
         }
     }
 
-    private val _speakers by lazy {
-        datastore.readSpeakers(conf).map { it.toSpeaker() }
-    }
```
martinbonnin, line 49: Does this remove the caching of the speaker? Last time I checked, having per-query caching of speakers did make a difference

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SessionsViewModel.kt
```diff
@@ -93,7 +92,7 @@ open class SessionsViewModel : KMMViewModel(), KoinComponent {
         val timeZone = sessionsData.config.timezone.toTimeZone()
         val sessionsMap =
             sessionsData.sessions.nodes.map { it.sessionDetails }.groupBy { it.startsAt.date }
-        val speakers = sessionsData.speakers.map { it.speakerDetails }
+        val speakers = sessionsData.speakers.nodes.map { it.speakerDetails }
```
martinbonnin, line 95: That's an issue, it's going to break apps currently in production. We're a real product now, we need to go through deprecation lifecycle.

BoD: Good point :) Just pushed a new field `speakersPage` and marked `speakers` as deprecated.

backend/service-graphql/src/main/kotlin/dev/johnoreilly/confetti/backend/graphql/DataStoreDataSource.kt
```diff
@@ -144,10 +161,30 @@ class DataStoreDataSource(private val conf: String, private val uid: String? = n
...
+        } else {
+            _speakers.indexOfFirst { it.id == after }
+                .also {
+                    if (it == -1) {
+                        return SpeakerConnection(emptyList(), PageInfo(null))
+                    }
+                } + 1
+        }
+        return SpeakerConnection(
+            nodes = _speakers.drop(drop).take(first),
```
martinbonnin, line 176: I always assumed `.take()` would throw if passed an argument longuer than the list. Turns out it's not. TIL!

martinbonnin (approved):
LGTM ! 

2 follow up things:
* We need to memoize sessions as well because with https://github.com/joreilly/Confetti/pull/418 we're now querying each session individually 👀 
* Would be nice to dump the generated schema somewhere so that we can double check there's no breaking change, a bit like we have with kotlin-binary-compatibility

PR #482, opened 2023-04-01
Fix A11y issue for loading placeholders

yschimke:
PLR Report

image


yschimke:
@joreilly there is a chance the placeholder screenshots become flaky, if so, I'll bump the tolerance to 100%

PR #494, opened 2023-04-02
Add Confetti project icon

yschimke:
Shows up when opening projects in Android Studio
image


yschimke:
@joreilly just converted from the png, but if you have a clean SVG we should update.

joreilly:
@martinbonnin has svg version I think

yschimke:
I might land and update, rather than coordinating on this PR.

martinbonnin:
Follow up there: https://github.com/joreilly/Confetti/pull/495

FWIW, the design sources are under `design`. Currently they are .sketch files cause some masking transformations do not work well in SVG

PR #498, opened 2023-04-02
Change Wear Theme to use safer route for Int colours.

yschimke:
I got some bizarre error with ColourSpace when applying a theme.  Fix to use safer Int path.  

The full representation is here https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/Color#representation

### DIFF
--- a/shared/src/commonMain/proto/wear.proto
+++ b/shared/src/commonMain/proto/wear.proto
@@ -4,22 +4,23 @@ option java_package = "dev.johnoreilly.confetti.wear.proto";
 option java_multiple_files = true;
 
 message Theme {
-  uint64 primary = 1;
-  uint64 primaryVariant = 2;
-  uint64 secondary = 3;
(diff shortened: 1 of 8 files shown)


yschimke:
@joreilly assuming thumbs up is a LGTM. Still building against iOS, but everything else passing, so will land and start fixing colour choices.

shared/src/commonMain/proto/wear.proto
```diff
@@ -4,22 +4,23 @@ option java_package = "dev.johnoreilly.confetti.wear.proto";
...
+  uint32 onSecondary = 9;
+  uint32 onBackground = 10;
+  uint32 onSurface = 11;
+  uint32 onSurfaceVariant = 12;
+  uint32 onError = 13;
 }
 
 message WearSettings {
-  Theme theme = 1;
+  reserved 1;
```
yschimke, line 23: This is because the encoding of Theme changed.

PR #499, opened 2023-04-02
Add past/upcoming filters to bookmarks

marcellogalhardo:
* Add past/upcoming tabs to bookmarks.
* Filter automatically based on the current time, refreshed every 5 minutes - it seemed reasonable, but we can fine-tune it as we wish.

Issues: the tab text color is different between sessions and bookmarks, we will need to fix it.

| Past | Upcoming |
|---|---|
| Screenshot_1680445258 | Screenshot_1680445261 |

### DIFF
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/utils/DateService.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/utils/DateService.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.isActive
 import kotlinx.datetime.LocalDateTime
 import kotlinx.datetime.TimeZone
 import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
 
 interface DateService {
     fun format(localDateTime: LocalDateTime, timeZone: TimeZone, format: String): String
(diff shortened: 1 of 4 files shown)


martinbonnin (approved):

yschimke:
Should Upcoming be the default and on the left, past be a secondary option?

martinbonnin:
+1 to default to upcoming

marcellogalhardo:
@yschimke the current page is `Upcoming`, it is defined in `val pagerState = rememberPagerState(initialPage = 1)` (I should have used `BookmarksTab.Upcoming.ordenal()` to make it explicit).

For the order point, I can change it as you folks prefer - no position from my side.

For context, my rationality is that we show the schedule as 13/04 and 14/04 so it felt intuitive to show the order as `Past` and `Upcoming`. Also, the KotlinConf app uses the same order.

yschimke:
@marcellogalhardo yep, missed this. No issue from me.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/utils/DateService.kt
```diff
@@ -1,11 +1,26 @@
...
 
 interface DateService {
     fun format(localDateTime: LocalDateTime, timeZone: TimeZone, format: String): String
 
     fun now(): LocalDateTime
-}
\ No newline at end of file
+}
+
+fun DateService.createCurrentLocalDateTimeFlow(delay: Duration = 5.minutes): Flow<LocalDateTime> =
```
yschimke, line 19: @marcellogalhardo I think this should be in the TZ of the conference, otherwise the comparisons will only work while you are at the conference?

PR #502, opened 2023-04-02
Fix colours for Mobile theme sync

yschimke:
Pretty sure the main bug was a false value in this line

```
val theme = colorScheme(false, false, false, context)
```


yschimke:
But I updated the TestFixtures with the correct colours, so screenshot tests show all possible defaults.

joreilly (approved):

PR #506, opened 2023-04-02
Mobile Auth on Wear

yschimke:


yschimke:
Abandoning to avoid needing mobile to do auth.

PR #511, opened 2023-04-03
Add navigation from session details to speaker details

### LINKED ISSUE #492: Add navigation from Session Details to Speaker
Clicking on the speaker photo on a session detail should navigate the user to the speaker details screen.

oas004:
In this PR I have added navigation to the speaker details of the speaker who is clicked on in the session details. 
This should resolve #492 

In the issue it says that if you click the image it should navigate. Does that make sense? Or should it be if you click the speaker detail section?

https://user-images.githubusercontent.com/54936943/229589890-81689668-af12-4902-aab4-c2db38c73942.mp4

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
@@ -37,22 +37,40 @@ import androidx.compose.ui.unit.dp
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import dev.johnoreilly.confetti.SessionDetailsViewModel
 import dev.johnoreilly.confetti.fragment.SessionDetails
+import dev.johnoreilly.confetti.speakerdetails.navigation.SpeakerDetailsKey
 import dev.johnoreilly.confetti.utils.format
 import org.koin.androidx.compose.getViewModel
 import java.time.format.DateTimeFormatter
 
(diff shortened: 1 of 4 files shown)


oas004:
> > Or should it be if you click the speaker detail section?
> 
> Would another option be to have row for each speaker clickable?

Not quite sure, but something like this could also be an option

https://user-images.githubusercontent.com/54936943/229591633-47cde2b6-59c9-46e0-8d2d-d3381a770ccb.mp4

oas004:
Maybe it is a bit strange to introduce this here when it is not like this anywhere else for ripples, but what do u think about rounding the corners on ripples like this? 😅 

https://user-images.githubusercontent.com/54936943/229595696-5def5a9c-f0a8-46be-8981-96734cbae187.mp4

joreilly:
hmm, no strong opinion TBH.....I guess the more standard the behaviour the better.

oas004:
Yeah good point! Let's keep it like the square one then :)

marcellogalhardo:
Nice job, thanks @oas004.

If I can add, I think the ripple should expand until the margins of the screen, and have a padding on the top and bottom. That would improve how the default ripple looks like (and remove this "pointed square" impression).

That can be corrected in a different PR, of course.

oas004:
Good point! Maybe that will make it look better! Will make a PR on that! :)

oas004:
Added a PR on it here: https://github.com/joreilly/Confetti/pull/529 :)

Issue #382, opened 2023-03-21
Tablet bookmarks


martinbonnin:
Add bookmarks to the tablet UI

oas004:
Is there a different design on bookmarks for tablet UI? Or is it just to add the destination? I tried that locally and this is how it looked : 

![image](https://user-images.githubusercontent.com/54936943/229624211-228d5470-0c7c-424d-b0f9-1fdd550039bd.png)

joreilly:
I think we can close this now?

PR #513, opened 2023-04-03
Add bookmarks to tablet session view

oas004:
Also I added the bookmark top level destination to the navigation rail. I'm not 100% sure about the design on this, so I'm leaving it a draft PR :) I also would like to add some robolectric tests to test the compose logic if that is okay? :) 

This relates to #382 

Here is what it looks like now:

image
image

If u have not signed in and clicks the bookmark
image

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListGridView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListGridView.kt
@@ -8,7 +8,6 @@ import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.horizontalScroll
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
@@ -95,7 +94,7 @@ fun SessionListGridView(
                         }
 
                         val timeInfoWidth = 90.dp
-                        val sessionInfoWidth = 240.dp
+                        val sessionInfoWidth = 280.dp
 
                         Column {
                             Row(
@@ -181,7 +180,7 @@ fun SessionGridRow(
             Surface(
                 modifier = Modifier
                     .width(sessionInfoWidth)
-                    .height(180.dp)
+                    .height(220.dp)
                     .padding(bottom = 16.dp)
                     .border(BorderStroke(1.dp, MaterialTheme.colorScheme.primary)),
                 color = MaterialTheme.colorScheme.secondaryContainer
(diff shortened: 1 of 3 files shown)


oas004:
Is it only wear that has robolectric tests? :) Can I add that for the Android app side?

joreilly:
That looks good....I think only concern I had before about putting in that position is if the name is very long (ran in to that elsewhere).  @martinbonnin what do you think?

martinbonnin:
What about something like hoverboard? https://hoverboard-master.web.app/schedule/2016-09-10

Screenshot 2023-04-04 at 10 55 47

Align title left and put the bookmark on the line below aligned right? We don't have tags/level for sessions but that might still work.

oas004:
Wow! That looks cool! I can try to implement that! 🙌

oas004:
Might be me that is doing this wrong, but should the title have some max line and truncate after that? If the title is very long I get an issue with the spacing for the speakers. Wdyt? In the design above they are truncating it, but that is a description so maybe that is a bit different 🤔 
This is the issue: 
image

oas004:
Yeah, I agree! :) Here I have increased the card size some. Wdyt? 

Here it is with the static CenterEnd align
image

Here it is in the column as an element between
image

oas004:
Is there a limit on how many users that can speak at one talk? :)

oas004:
So something like this if it is more than 3? In this case it had a lot of padding since the title is not very large 🤔 
image

martinbonnin:
Agree, feels weird. Maybe 2 x 2 + ellipsis?

oas004:
Maybe this makes it better, as a FlowRow with 2 items max in each row 🤔 
image

martinbonnin:
> Maybe this makes it better, as a FlowRow with 2 items max in each row 🤔

I was thinking with names. 

```
[pic] name                  [pic] name2
[pic] other long na..   [pic] name3
```

But design review over GitHub is hard! Feel free to choose the one you like best and then we can iterate during KotlinConf :)

oas004:
> Agree, feels weird. Maybe 2 x 2 + ellipsis?

Yeah, I think that can work 🤔

oas004:
> > Maybe this makes it better, as a FlowRow with 2 items max in each row 🤔
> 
> I was thinking with names.
> 
> ```
> [pic] name                  [pic] name2
> [pic] other long na..   [pic] name3
> ```
> 
> But design review over GitHub is hard! Feel free to choose the one you like best and then we can iterate during KotlinConf :)

Yeah I agree! Sadly I didn't order tickets in time for that :( But I'll be watching the stream! 🙌

martinbonnin:
Oh no! I was looking forward to catchup! Well, just in case you want to visit Paris in Spring 🌸 , [Android Makers](https://androidmakers.droidcon.com/tickets/) is a couple of weeks later and still has tickets :)

oas004:
Yeah that would have definitely been cool! 🤩

oas004:
I would like to add Robolectric tests for this PR. Is it only wear that has any robolectric tests? Can I configure that for Android as well? :)

martinbonnin:
Yup, I think there's nothing on mobile yet, go for it.

PR #525, opened 2023-04-04
Show time and room in session details

### LINKED ISSUE #487: Show session time and room in the session details screen

oas004:
Added the time and room for to be displayed in the session details. 

This should resolve #487 

I used `labelLarge` for these texts. Not quite sure if that is correct. According to M3 docs, you can use that for small texts in text bodys like captions (https://m3.material.io/styles/typography/applying-type#af6eb002-9cbb-4b64-bce6-1315d2252364). I can change it up if that is preferred. Maybe it's a bit small. Here is how it looks

Android Phone

image

Android Tablet
image


joreilly:
Small thing, and could be follow on PR, but might be good to show date as well as time in this view.

oas004:
I can add that! Is there a specific format that should be used? :)

oas004:
Didn't know there was an official KotlinConf app 😅 But that looks nice! Updated with that time formatting in [91d41b7](https://github.com/joreilly/Confetti/pull/525/commits/91d41b70dae8d137fbe26d161cdf3895b96e8b95) :)

joreilly:
I think I still prefer seeing the title first.

PR #529, opened 2023-04-05
Improve ripple on speaker view on session details

oas004:
Following up on comment from @marcellogalhardo (https://github.com/joreilly/Confetti/pull/511#issuecomment-1497821209).  Trying to improve the ripple effect so that it is going all the way to the sides horizontally. This was to remove the pointy square feeling.

Now it looks like this:

https://user-images.githubusercontent.com/54936943/230162005-2179d00c-2f56-4bbe-8f9e-7a65e409c049.mp4

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
@@ -96,15 +96,17 @@ fun SessionDetailView(
         }
     ) {
         Column(modifier = Modifier.padding(it)) {
+            val horizontalPadding = 16.dp
             session?.let { session ->
                 Column(
                     modifier = Modifier
                         .fillMaxWidth()
-                        .padding(horizontal = 16.dp, vertical = 8.dp)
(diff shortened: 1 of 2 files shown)


androidApp/src/main/java/dev/johnoreilly/confetti/sessiondetails/SessionDetailsView.kt
```diff
@@ -113,13 +115,15 @@ fun SessionDetailView(
...
                     Text(
+                        modifier = Modifier.padding(horizontal = horizontalPadding),
                         text = session.startsAt.toTimeString(session.endsAt),
                         color = MaterialTheme.colorScheme.onSurface,
                         style = MaterialTheme.typography.labelLarge
                     )
 
                     session.room?.name?.let { roomName ->
                         Text(
+                            modifier = Modifier.padding(horizontal = horizontalPadding),
```
oas004, line 126: Duplicating these does not feel quite correct, but the issue here is that I need to set the click modifier before the padding so that the padding does not limit the ripple. So I have to move the padding out of the column. Is there a better way to do this? 🤔

PR #537, opened 2023-04-06
Enable predictive back gesture

StylianosGakis:
We're using androidx.navigation already, so we can opt in and get this feature for free for those phones that have opted into this flag. What do you think?

| Before | After |
| --- | --- |
| https://user-images.githubusercontent.com/44558292/230495796-2a4365e1-3492-44ac-8fa8-af14e731625b.mp4 | https://user-images.githubusercontent.com/44558292/230495843-76716ed5-eeff-4648-b55a-67fc639d6625.mp4 |

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -16,7 +16,8 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:label="@string/app_name"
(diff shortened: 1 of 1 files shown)


StylianosGakis:
Btw not sure how to make the video play inline in the PR view 🤔

martinbonnin:
> Btw not sure how to make the video play inline in the PR view

I've used gif before but maybe Github supports video now, not sure.

joreilly:
very cool!

PR #539, opened 2023-04-06
Fix session screen having some unnecessary bottom padding

StylianosGakis:
Make use of window insets instead, to specifically only pad that much for cases like tablets, while still making sure that in phone form factors the insets are not applied two times.

How it looked like before:
landscape_too_much_bottom_padding

Changing from padding(80) to using insets initially changes it to this (Got some red border to indicate where the content should be):
That extra spacing is from the window insets for the navigation bar being applied again
before_consuming_the_bottom_bar_paddings

The `consumedInsetsModifier` I added makes sure that we inform the content that the insets are already consumed. This does not happen automatically in this case since `ConfettiBottomBar` and `content()` are two separate items, and the way that the inset modifiers work is that they inform items that are children of wherever they are applied. Since now they're not inside each other but in a column side by side, we need to do this ourselves.

And finally we get this result:
| Phone | Tablet |
| --- | --- |
| after_fix | after_fix_tablet |

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListGridView.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/sessions/SessionListGridView.kt
@@ -13,12 +13,17 @@ import androidx.compose.foundation.layout.ExperimentalLayoutApi
 import androidx.compose.foundation.layout.FlowColumn
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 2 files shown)


joreilly:
Thanks for this.  I had wondered whether this was right layout for phone in landscape (as opposed to tablet) but great to have this change anyway.

joreilly:
I think there might be flaw in existing logic here.....if we're showing grid layout like this we should at least also be showing nav rail instead of bottom bar....

Issue #533, opened 2023-04-06
Indicate lightning talks in the UI


joreilly:
Can be inferred we believe from the talk duration....see following for example https://kotlinconf.com/schedule/april-13/

<img width="785" alt="Screenshot 2023-04-06 at 11 50 20" src="https://user-images.githubusercontent.com/6302/230355920-af50ed32-84e9-49bf-988f-bc9d82161816.png">

yschimke:
Shouldn't we put this sort of logic behind the server? perhaps with a more generic concept?

StylianosGakis:
Aha, some sort of extra tag, which would support arbitrary text in there I presume?
I see that in the website they also have another similar tag which says "AWS lab", with a different icon. So yeah that'd probably be a good idea.

A question for this one, how would we be sending the icon to the clients from the server? Does Confetti right now have a place where such icons are hosted so that we can send the URL from the backend to the apps?

joreilly:
@martinbonnin what do you think about having this in the backend?  We can probably close this particular issue for now and create new one if we think that's something we'd do

martinbonnin:
Agreed, it'd be interesting to investigate some kind of server-driven rendering for this. There is already "session.tags" but it's very loosely coupled. Let's close this issue and revisit for Devoxx "quickies" ;)

PR #540, opened 2023-04-07
Indicate lightning talks

StylianosGakis:
Partially addresses #533

Used `Icons.Default.Bolt` as a simple way to get something resembling what's in the website as discussed here https://kotlinlang.slack.com/archives/C051P2HUVKP/p1680724416100319. If we want the actual same icon as the website I wasn't sure how you'd like to bring that in so I didn't do it.

### Android
| Phone | Tablet |
| --- | --- |
| phone_lightning | tablet_lightning |

Plus a little change in the modifier order so that the click interaction in the card spans across the entire card.

### iOS

| Light mode | Dark mode |
| --- | --- |
| simulator_screenshot_F711D0FC-AB8C-48AC-835B-7B8E7468D12D | simulator_screenshot_2 |

### Room for improvement:
The iOS implementation probably can use some love, I've never really written SwiftUI code before 😅 Also didn't add the lightning as I wanted to hear which one you'd like to use there, as it may change for Android too.
Get opinions about the color decision of using `primaryContainer`, especially in the big screen where the primaryContainer isn't as nice as it is on the small screen which is on top of just the background color.


joreilly:
nice!

PR #541, opened 2023-04-07
update logic for when to show nav rail

joreilly:
shows nav rail now on phone in landscape

Screenshot_20230407_101402


StylianosGakis (approved):
That looks so much nicer now 🤩

PR #544, opened 2023-04-07
Add notification for upcoming bookmarked sessions

marcellogalhardo:
* Add notification for upcoming bookmarked sessions.
* We do not handle permissions, a user needs to manually set the permission;
* Clicking on the notification does NOT open the notification session details in the app;
* Includes an experimental flag that allows disabling the feature. The default value is disabled.

Pending:
* We should request the user's permission in the app;
* Clicking on the notification should open session details;
* We should optimize the bookmarked sessions query, so we do not need to load all conferences (cc: @martinbonnin).

| Notifications | Collapsed | Expanded | Channel | Experimental Settings |
|---|---|---|---|---|
| Screenshot_1680889806 | Screenshot_1680867551 | Screenshot_1680867557 | Screenshot_1680867712 | Screenshot_1680889966 |


shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        val user = authentication.currentUser.value
+
+        val conferences = repository.conferences(FETCH_POLICY)
+            .data
+            ?.conferences
+            .orEmpty()
+
+        val bookmarks = conferences.map { conference ->
+            repository.bookmarks(
```
marcellogalhardo, line 43: I do not know much about Confetti's GraphQL setup but I suspect the code here is sub-optimal and forces all conferences to be loaded. Maybe @yschimke or @martinbonnin can help me here?

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationWorker.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        fun startWorkRequest(workManager: WorkManager) {
+            workManager.enqueueUniquePeriodicWork(
+                TAG,
+                ExistingPeriodicWorkPolicy.UPDATE,
+                createPeriodicWorkRequest(),
+            )
+        }
+
+        private fun createPeriodicWorkRequest(): PeriodicWorkRequest =
```
yschimke, line 37: Running this periodically seems wasteful. Instead could we use the bookmark toggle action to create a once off action that will send the notification at that point. Use the daily job to sync them.

Longer term, we may want to consider sending the notifications via Firebase from the server.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val repository: ConfettiRepository,
+    private val dateService: DateService,
+    private val notificationManager: NotificationManagerCompat,
+    private val authentication: Authentication,
+) {
+
+    suspend fun sendNotification() {
+        Log.d("marcello", "notify")
+
+        val user = authentication.currentUser.value
```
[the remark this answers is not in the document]
marcellogalhardo: Done.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -0,0 +1,158 @@
...
+
+        val conferences = repository.conferences(FETCH_POLICY)
+            .data
+            ?.conferences
+            .orEmpty()
+
+        val bookmarks = conferences.map { conference ->
```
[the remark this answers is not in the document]
marcellogalhardo: Done.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val notificationManager: NotificationManagerCompat,
+    private val authentication: Authentication,
+) {
+
+    suspend fun sendNotification() {
+        Log.d("marcello", "notify")
+
+        val user = authentication.currentUser.value
+
+        val conferences = repository.conferences(FETCH_POLICY)
```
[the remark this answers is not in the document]
marcellogalhardo: Done.

yschimke (approved):
LGTM

I think other changes are way bigger, and given mobiles have big battery, polling for notifications is probably fine.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationWorker.kt
marcellogalhardo: > Running this periodically seems wasteful.

Agreed.

> Instead could we use the bookmark toggle action to create a once off action that will send the notification at that point.

Do you mean, whenever the user "adds or removes a new bookmark" we trigger the notifications? If yes, I think that may not be that helpful. The use case I have in mind is:
- I add bookmarks for the conference during the week.
- On the event day, while I'm distracted, the app would notify me a given session is to start so I can go to the room.

FYI: The photo above shows all bookmarked sessions because I removed the filter during my tests but in the code, we [check only the upcoming sessions in a range of up to 15 minutes](https://github.com/joreilly/Confetti/pull/544/files#diff-69b1d1f49e0d1bd139663b6ae3805ea0c9c2d2f8b047a08d104bbfefead356a9R65-R68).

In other words, I would receive notifications throughout the day whenever a session is 15 minutes about to start without requiring me to interact with the app as the room is visible.

> Longer term, we may want to consider sending the notifications via Firebase from the server.

Agreed.

---

Let me know what you think!

yschimke: > Do you mean, whenever the user "adds or removes a new bookmark" we trigger the notifications? If yes, I think that may not be that helpful

No, I meant we don't need to run a job every 15 minutes to create notifications for the next 15 minutes.

We can create jobs scheduled to run when the notification should show. Update these jobs users add or remove favourites.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -0,0 +1,162 @@
...
+        val bookmarkedSessions = sessions.filter { session ->
+            bookmarks.contains(session.id)
+        }
+
+        val intervalRange = createIntervalRange()
+        val sessionsToNotify = bookmarkedSessions.filter { session ->
+            session.startsAt in intervalRange
+        }
+
+        if (sessionsToNotify.isNotEmpty()) {
```
yschimke, line 83: Since you have autoCancel, what's stopping this from recreating notifications the user has dismissed?

yschimke (changes requested):
Can you confirm the behaviour after user dismisses a notification?

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
marcellogalhardo: There is nothing preventing recreation but that should not happen because:
1. The period work only runs each 15 minutes.
2. We only show the sessions "starting" in the next 15 minutes.

Not the most elegant solution. tho.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationWorker.kt
marcellogalhardo: Ahhh, I see. That makes a lot of sense, we should definitely do that.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
marcellogalhardo: I will double-check the logic but be sure I didn't miss anything.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
marcellogalhardo: We are only using the current conference. Stale.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
marcellogalhardo: Hey @yschimke I checked and it is working as expected per message [here](https://github.com/joreilly/Confetti/pull/544#discussion_r1160686617).

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
```diff
@@ -70,13 +68,19 @@ class SessionNotificationSender(
         }
 
         val intervalRange = createIntervalRange()
-        val sessionsToNotify = bookmarkedSessions.filter { session ->
+        var sessionsToNotify = bookmarkedSessions.filter { session ->
```
yschimke, line 71: why var?

yschimke (approved):
Seems safe if behind a flag.

androidApp/src/main/java/dev/johnoreilly/confetti/ConfettiApplication.kt
```diff
@@ -46,6 +49,15 @@ class ConfettiApplication : Application(), ImageLoaderFactory {
 
         val workManager = get<WorkManager>()
         setupDailyRefresh(workManager)
-        SessionNotificationWorker.startWorkRequest(workManager)
+
+        ProcessLifecycleOwner.get().lifecycleScope.launch {
```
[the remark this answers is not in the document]
marcellogalhardo: It seems the PR was merged, will try to add the comment in a follow-up PR.

shared/src/androidMain/kotlin/dev/johnoreilly/confetti/work/SessionNotificationSender.kt
marcellogalhardo: For absolutely no reason. 🙃  Probably a left-over from my experiments. Will remove it in a follow-up PR.

PR #549, opened 2023-04-07
Add bookmarks to Wear app

yschimke:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -86,11 +86,11 @@ class ConfettiRepository : KoinComponent {
         uid: String?,
         tokenProvider: TokenProvider?,
         fetchPolicy: FetchPolicy
-    ): ApolloResponse<GetBookmarksQuery.Data> =
+    ): Flow<ApolloResponse<GetBookmarksQuery.Data>> =
         apolloClientCache.getClient(conference, uid).query(GetBookmarksQuery())
             .tokenProvider(tokenProvider)
             .fetchPolicy(fetchPolicy)
-            .execute()
+            .toFlow()
```
yschimke, line 93: This can't be execute if we take fetchPolicy in as a parameter.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -150,11 +150,15 @@ class ConfettiRepository : KoinComponent {
         }
     }
 
-    suspend fun sessionDetails(
+    fun sessionDetails(
```
yschimke, line 153: It feels like we don't have a common plan for how to implement these.  A similar issue happens with Room, for suspend functions vs Flow.

But especially here, it would be nice if we could hide away differences between CacheFirst, CandAndNetwork, NetworkFirst.

I'm unclear what happens when CacheAndNetwork, gets a valid cache result, but then fails on network.  Does it silently drop it, so the cache result stands? Or show an error?

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SpeakerDetailsViewModel.kt
```diff
@@ -27,7 +27,7 @@ class SpeakerDetailsViewModel(
 
     val speaker: StateFlow<UiState> = flow {
         // FixMe: add .speaker(id)
-        val response = repository.conferenceData(conference = conference, FetchPolicy.CacheOnly)
+        val response = repository.conferenceData(conference = conference, FetchPolicy.CacheFirst)
```
yschimke, line 30: CacheFirst seems safer and more like what we want than CacheOnly.

wearApp/snapshot/ConferenceHomeScreenTest/conferenceHomeScreen_Colors_Material__reference.png
yschimke, line 1: On the pages that are information focused, instead of action focused.  I decided to make this more muted.

```
                    chipType = StandardChipType.Secondary
```

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
martinbonnin: >  what happens when CacheAndNetwork, gets a valid cache result, but then fails on network

It emits 2 `ApolloResponse`:

* First one with `apolloResponse.exception == null`
* Seconf one with `apolloResponse.exception != null`

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
martinbonnin: Agreed. `execute()` is a bit awkward because of `CacheAndNetwork` that emits twice. There is a bit of brain muscle around it though so it's hard to change at this point.

shared/src/mobileMain/kotlin/dev/johnoreilly/confetti/SpeakerDetailsViewModel.kt
martinbonnin: +1. That'll fix a deepling to a speaker details if the conference has never been opened before.
Also we should get only the speaker instead of retrieving all the conference data

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
yschimke: What about the data?

Should we ignore that second one? assume it's just network is down?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
yschimke: I might add a simple one so it's hidden from clients.

yschimke: Probably an extension function to ApolloCall?

martinbonnin: Not sure I follow. Soemthing like this?

```
ApolloCall.cacheAndNework()
```

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
martinbonnin: > What about the data?

* First one with `apolloResponse.data != null`
* Second one with `apolloResponse.data == null`

> Should we ignore that second one? assume it's just network is down?

I would ignore it yes. And rely on pull to refresh if the user really want latest data.

To me:

* initial load with `CacheAndNetwork` means: display something as soon as possible, update in the background
* pull to refresh means: I always want fresh network results (or an error else)

The more I think about it, the less I like `CacheAndNetwork` though because it updates the data without user action which might feel a bit weird.

yschimke: > The more I think about it, the less I like `CacheAndNetwork` though because it updates the data without user action which might feel a bit weird.

I like that, but it's a change for mobile.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
yschimke: Not sure, I think I want `ApolloCall.asStateFlow(fetchPolicy): StateFlow<Result<Data>>` that hides all errors when it has any results to show. And a failed form if both network and cache failed.

martinbonnin: Hiding errors is a bit risky. There's some background there where we did that for `.watch()` and some users required access to the individual errors.

yschimke: What about Flow<ApolloResponse> but it doesn't go from a a non null data, to null?

martinbonnin: Maybe yea. In all cases, I'd proceed with caution there as we've been burnt by this before (one instance [there](https://github.com/apollographql/apollo-kotlin/issues/3976)). Maybe keep the current `toFlow()` that emits everything and add helper functions on top of it. If it's `StateFlow` it can be one layer of abstraction above or so.

PR #556, opened 2023-04-08
Add a UIState Flow

yschimke:
An attempt at a consistent UI approach for turning queries into StateFlow.

### DIFF
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
@@ -109,7 +109,7 @@ class ConfettiRepository : KoinComponent {
     /**
      * ignores all errors unless the Flow terminates without an emission in which case it will emit an error
      */
-    private  fun <D: Operation.Data>  Flow<ApolloResponse<D>>.itemsOrError(operation: Operation<D>): Flow<ApolloResponse<D>> {
+    private fun <D: Operation.Data>  Flow<ApolloResponse<D>>.itemsOrError(operation: Operation<D>): Flow<ApolloResponse<D>> {
         var hasValidResponse = false
         return this.onEach {
             if (it.data != null) {
@@ -124,7 +124,7 @@ class ConfettiRepository : KoinComponent {
         }
     }
 
-    suspend fun bookmarks(
+    fun bookmarks(
         conference: String,
         uid: String?,
         tokenProvider: TokenProvider?,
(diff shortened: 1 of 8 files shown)


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ConfettiRepository.kt
```diff
@@ -195,14 +195,21 @@ class ConfettiRepository : KoinComponent {
         }
     }
 
-    fun sessionDetails(
+    fun sessionDetailsQuery(
```
yschimke, line 198: In this particular future world, all repositories are only defined in terms of query functions returning ApolloCall.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/utils/ClientQuery.kt
```diff
@@ -0,0 +1,38 @@
...
+        if (next.data != null) {
+            QueryResult.Success(mapper(next.data!!))
+        } else {
+            previous
+        }
+    }
+}
+
+sealed interface QueryResult<out T> {
+    data class Error(
```
yschimke, line 29: TODO, one more change, to enforce the error case.

yschimke:
@martinbonnin this is the UI module extension I'd love to exist in apollo, maybe also with an option to also watch after the initial requests.

Happy to move this to Wear module if we don't think it's something we want overall.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/utils/ClientQuery.kt
```diff
@@ -0,0 +1,42 @@
...
+        val apolloException = next.exception
+
+        if (next.hasErrors() || apolloException != null) {
+            onError(next.errors.orEmpty(), apolloException)
+        }
+
+        if (next.data != null) {
+            QueryResult.Success(mapper(next.data!!))
+        } else if (apolloException != null && previous is QueryResult.Loading) {
+            QueryResult.Error(apolloException)
```
martinbonnin, line 25: That doens't work with `CacheAndNetwork` and your first item is an error. It will return Error here while I think you want to ignore it?

yschimke: Yep, it's probably the type of thing that a very opinionated screen might want to decide.  But mostly here, I'm ok with

Loading -> Error -> Success

if it's uncommon. If success takes 3-4 seconds over the network, do you want it to show loading, or an error for a short while?

What do you suggest?

PR #557, opened 2023-04-08
Add a splash screen and simplify home screen params

yschimke:


yschimke:
Nothing major changing in shared code, and copies the approach from @martinbonnin on mobile.

PR #558, opened 2023-04-08
Use new bookmarks query

yschimke:


yschimke:
Will refactor later, once https://github.com/joreilly/Confetti/pull/556 lands

PR #573, opened 2023-04-09
Use toUiState for the conferences screen.

yschimke:


wearApp/src/main/java/dev/johnoreilly/confetti/wear/conferences/ConferencesViewModel.kt
```diff
@@ -3,45 +3,31 @@ package dev.johnoreilly.confetti.wear.conferences
...
 
 class ConferencesViewModel(
     private val tileUpdater: TileUpdater,
     private val complicationUpdater: ComplicationUpdater,
     private val repository: ConfettiRepository,
-    private val workManager: WorkManager,
     private val refresh: ConferenceRefresh
 ) : ViewModel() {
-    val conferenceList: StateFlow<ConferencesUiState> = flow {
-        var hasConferences = false
```
yschimke, line 25: My hope is that ViewModels don't need to deal with this complexity, but can instead just us this function with different fetchModes.  I'm also changing this to CacheFirst, which I think we agreed.

yschimke:
Martin it out, so will land and address comments later.

PR #575, opened 2023-04-09
Change the approach to home screen

yschimke:
Home screen now reads from the conference setting, and shows the screen, reusing the content from the conference home screen.

I think we could probably merge these into a single screen, but it seems useful to have these separate. Let's see.


yschimke:
@joreilly going to land and do a little bit more testing on top of this.

PR #576, opened 2023-04-09
Use UiState for Home Screen

yschimke:


yschimke:
Enjoy your ride, I'll fix up any comments tonight.

PR #583, opened 2023-04-10
Improve the Wear refresh logic

yschimke:
Was trying to run too often, would be running every 15 minutes while charging.
Don't run when user selects a conference, as we are about to do that anyway.


androidApp/src/main/java/dev/johnoreilly/confetti/di/AppModule.kt
```diff
@@ -59,5 +57,4 @@ val appModule = module {
     single<WearSettingsSync> {
         WearSettingsSync(get(), get(), androidContext(), get())
     }
-    workerOf(::RefreshWorker)
```
yschimke, line 62: We aren't using on android yet, so omit.

joreilly (approved):

PR #592, opened 2023-04-11
Remove splash destination, add deep links and fix bottom navigation

### LINKED ISSUE #551: Consider handling bottom nav (and NavRail) navigation as top-level destinations launched as singleTop who have their own backstack
The bottom nav might feel a bit more intuitive to use when adding new destinations on top doesn't make the backstack list grow indefinitely as more destinations are pressed.

Current implementation looks like this:

| Confetti | NiA |
| --- | --- |
| https://user-images.githubusercontent.com/44558292/230629586-f7730731-f389-4a74-a3e2-7e154149952e.mp4 | https://user-images.githubusercontent.com/44558292/230629364-a5254591-e3e5-4298-99d1-5da48128d5b0.mp4 |

We can try to make Confetti function the same way as NiA does with this.

StylianosGakis:
Adds a deep link to the speakers screen and to the sessions screen. Adds the "confetti/confetti" scheme to the manifest just like the wear app does too.

Fix #551 by making bottom nav destinations save their state and restore it when returning back to them. Also ensures that when navigating between them, the entire backstack is not popped, but the start destination stays in the backstack, just like NiA handles this

Super important to note here that:
We seem to have to clear the backstack of the bottom nav top level destinations when we go away from that flow and go into the conferences screen. If we do not do that, the previously saved destinations that we are saving the state of as we pop them off will take precedence over the new routes we pass in. This means that they will be referencing the old conference argument, despite us sending in a new one, which seems to be completely ignored. So if we try to go to "speakers/kotlinconf2023" but we've before that popped a "spekers/fosdem2023" destination and saved the state, it will ignore our "kotlinconf2023" parameter and still have "fosdem2023" in its backstack parameters.
We do this inside `onSwitchConference` for now

Adds official splash screen support and add a SplashReadyStatus class to hold the state. The top level destinations can use this to report when they are ready to get rid of it. In the future this can be postponed even further to make sure that the sessions screen waits not only until it is ready to load the data, but after it has already started displaying the first batch of data.
For now the first two screens that one can see when opening the app are either the conferences destinations or the sessions destinations, and both of these report when they are ready to show. If we are opening an app using a deep link, we simply do not postpone getting rid of the splash screen in the first place, and let those screens show immediately.


androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -19,18 +22,31 @@ import dev.johnoreilly.confetti.settings.DarkThemeConfig
...
 class MainActivity : ComponentActivity() {
 
+    private val confettiRepository: ConfettiRepository by inject()
+    private val splashReadyStatus: SplashReadyStatus by inject()
+
     override fun onCreate(savedInstanceState: Bundle?) {
+        val splashScreen = installSplashScreen()
         super.onCreate(savedInstanceState)
 
+        if (intent.data == null) { // If we're deep linking anywhere, don't show splash
```
StylianosGakis, line 43: I hope this is an adequate indicator of "we're now handling a deep link"

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -64,28 +81,50 @@ class MainActivity : ComponentActivity() {
...
+     * Now as I see it we only get the already decoded Uri so I don't know how to get the ID in a
+     * better way.
+     */
+    private fun Uri.findConferenceParameter(): String? {
+        return pathSegments.also { Log.d("Stelios", "segmn$it") }
+            .dropWhile<String?> { segment -> segment != "sessions" && segment != "speakers" }
+            .drop(1) // Also drop the "sessions" or the "speakers" segment
+            .firstOrNull()
+    }
 }
```
StylianosGakis, line 111: This whole section relies on the urls looking exactly like this as I am describing here. If there's a better way to do this please do let me know. Went off of our discussion here https://kotlinlang.slack.com/archives/C051P2HUVKP/p1681162362863849?thread_ts=1680994152.780399&cid=C051P2HUVKP

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/HomeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+    confettiRepository: ConfettiRepository,
+    splashReadyStatus: SplashReadyStatus,
+) : ViewModel() {
+    private val conferenceParam: String? =
+        SessionsKey.conferenceFromNavArgs(savedStateHandle)?.ifEmpty { null }
+
+    private val conferenceFlow: Flow<String> = flow {
+        if (conferenceParam != null) {
+            // Set the conference selected if we've deep linked into a specific conference
+            confettiRepository.setConference(conferenceParam)
```
StylianosGakis, line 31: This is the part that makes sure that if we've deep linked into a specific conference we also select it in the repository.
This made sense to me so that if you deep link into one, then close the app, then open it again I'd suspect you'd want to see what you saw last time too, not what you pressed the time *before* the last time you had the app open.

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/HomeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+                HomeUiState.Conference(conference)
+            } else {
+                HomeUiState.NoConference
+            }
+        }
+        .onEach { uiState ->
+            if (uiState is HomeUiState.Conference) {
+                splashReadyStatus.reportIsReady()
+            }
+        }
```
StylianosGakis, line 49: This reports to the splash screen that we are ready to show the UI. This can be postponed even further however to inside the sessions screen and *after* the first GQL request is done so that the first frame already shows content instead of seeing a loading spinner for a brief moment. Can be addressed in a future PR however.

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/HomeViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+            viewModelScope,
+            SharingStarted.WhileSubscribed(5.seconds),
+            HomeUiState.Loading
+        )
+}
+
+sealed interface HomeUiState {
+    object Loading : HomeUiState
+    data class Conference(val conference: String) : HomeUiState
+    object NoConference : HomeUiState
```
StylianosGakis, line 60: This state is the one that skips showing the sessions screen completely (the splash screen is still showing here mind you), and opens the conferences screen as the single item in the backstack, which itself can report when it's ready to be shown

androidApp/src/main/java/dev/johnoreilly/confetti/sessions/navigation/SessionsDestination.kt
```diff
@@ -30,22 +43,41 @@ fun NavGraphBuilder.sessionsGraph(
...
+                    conference = homeUiState.conference,
+                    appState = appState,
+                    navigateToSession = navigateToSession,
+                    onSignOut = onSignOut,
+                    navigateToSignIn = navigateToSignIn,
+                    onSwitchConferenceSelected = onSwitchConferenceSelected,
+                )
+            }
+        }
     }
```
StylianosGakis, line 82: This "extra layer" sort of was added in the sessions destination to decide on if it show the sessions route or to go to the conferences destination, without having to add a third helper destination to support this.

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiApp.kt
```diff
@@ -48,22 +58,48 @@ fun ConfettiApp(
...
-            appState.navigateToTopLevelDestination(ConferencesKey.route)
+            coroutineScope.launch {
+                confettiRepository.setConference(AppSettings.CONFERENCE_NOT_SET)
+                TopLevelDestination.values.forEach { destination ->
+                    // We *must* do this, else, the previously saved destinations will take
+                    // precedence over the new routes we pass in. This means that they will be
+                    // referencing the old conference argument, despite us sending in a new one,
+                    // which seems to be completely ignored.
+                    navController.clearBackStack(destination.routePattern)
+                }
```
StylianosGakis, line 69: This section handles the backstack clearing point that I make in this PR's comment.

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -64,28 +81,50 @@ class MainActivity : ComponentActivity() {
...
+     * or more generally
+     * `whatever/[sessions|speakers]/{conference}/whatever
+     *
+     * Should be doing some more general checks here or find a way to find what the original
+     * URI template was like, so we can find what is in the position of {conference} for example.
+     * Now as I see it we only get the already decoded Uri so I don't know how to get the ID in a
+     * better way.
+     */
+    private fun Uri.findConferenceParameter(): String? {
+        return pathSegments.also { Log.d("Stelios", "segmn$it") }
```
yschimke, line 106: Do we need this logging here?

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -19,18 +22,31 @@ import dev.johnoreilly.confetti.settings.DarkThemeConfig
...
+    private val confettiRepository: ConfettiRepository by inject()
+    private val splashReadyStatus: SplashReadyStatus by inject()
+
     override fun onCreate(savedInstanceState: Bundle?) {
+        val splashScreen = installSplashScreen()
         super.onCreate(savedInstanceState)
 
+        if (intent.data == null) { // If we're deep linking anywhere, don't show splash
+            // Keep splash screen on until we get the first info on which conference we've chosen
+            splashScreen.setKeepOnScreenCondition {
```
yschimke, line 45: This is nice!

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
yschimke, line 43: What about doing this based on the nav stack?  If you pull up the nav controller to the activity, then you can probably just check if we are on the home screen.

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -64,28 +81,50 @@ class MainActivity : ComponentActivity() {
...
+                        defaultConferenceParameter = defaultConferenceParameter,
                     )
                 }
             }
         }
     }
+
+    /**
+     * This relies on our deep link looking either like
+     * `confetti://confetti/sessions/{conference}`
```
yschimke, line 94: if we want to do this, we should probably pull up our navigation code like SessionDestination to shared code, and handle it the same across wear and mobile.

I can follow up with this.

androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiApp.kt
yschimke: This shouldn't be needed if the navOptions below are working correctly.

joreilly:
@StylianosGakis had you any more changes to make here or are you happy for it to be merged?

StylianosGakis:
Yeah there were a bunch of things that Yuri mentioned that would need to be addressed before the merge, so let's keep it open for a bit more. I haven't found the time to sit and see how I would address everything, will try to do whenever I find some extra time 😊

StylianosGakis:
Wanted to see how this https://github.com/joreilly/Confetti/pull/688#issuecomment-1555106309 would go after I saw this discussion starting. This can be closed in favor of that now.

PR #597, opened 2023-04-12
[For discussion] Use secondary chip for conferences screen

yschimke:
Advice from a UX colleague.

thoughts?


joreilly:
Looks nice!

joreilly:
Looks nice I think....was there any particular thing that the UX person mentioned about what was there already?

PR #598, opened 2023-04-12
[After kotlinconf] Update Speaker screen design

yschimke:


shared/src/commonMain/graphql/extra.graphqls
```diff
@@ -1,4 +1,6 @@
 extend type Session @typePolicy(keyFields: "id")
 extend type Bookmarks @typePolicy(keyFields: "id")
+extend type Speaker @typePolicy(keyFields: "id")
```
yschimke, line 3: @martinbonnin I made an attempt here, but not sure I made the right entries.

martinbonnin (approved):

PR #602, opened 2023-04-12
Propagate theme values to screenshot test function

### LINKED ISSUE #568: Set correct mobile background for RNG tests on AndroidApp  side
Following up https://github.com/joreilly/Confetti/pull/528#discussion_r1161158192 to remember to change the background color that is set on the androidApp side of the RNG tests

oas004:
Need this to be able to set the theme colors correctly

This resolves #568 

My plan is further to iterate on this API to enable multiple screenshot tests for each theme provided.  Will experiment a bit with that, so this is not the final API.  This just enables me to add screenshots for different theme values.


androidApp/snapshot/SpeakerDetailsScTest/screenshotDetailsA11y.png
yschimke, line 1: Do we expect a transparent background here?

androidApp/src/test/kotlin/dev/johnoreilly/confetti/SpeakerDetailsScTest.kt
```diff
@@ -26,19 +25,19 @@ class SpeakerDetailsScTest : BaseScreenshotTest(false, a11yEnabled = true) {
             sessions = emptyList(),
             socials = emptyList()
         )
-        screenshotTestRule.takeScreenshot(
+        takeScreenshot(
```
yschimke, line 28: Can we add one non A11y test which won't obscure the theme colours?

yschimke (approved):

androidApp/snapshot/SpeakerDetailsScTest/screenshotDetailsA11y.png
oas004: I was wondering about this as well. Would it make sense to have a public function in the theme that exposes the colors so that we can call `ConfettiTheme.colors.background` instead here? :) I think the background color would be correct depending on the theme that is provided.  Or do we do this in another way in Confetti? 🤔 

I saw that Jetsnack does that [here](https://github.com/android/compose-samples/blob/c62ff8483d111f06ac51946272278d6348b487f0/Jetsnack/app/src/main/java/com/example/jetsnack/ui/theme/Theme.kt#L110) and uses it [here](https://github.com/android/compose-samples/blob/c62ff8483d111f06ac51946272278d6348b487f0/Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/FilterScreen.kt#L211)

oas004: Aha okay, then I will add that instead :) Thanks!

androidApp/snapshot/SpeakerDetailsScA11yTest/screenshotDetailsA11y.png
yschimke, line 1: This makes sense.  The transparent background in screenshot tests is defaulted to the device background, but with the checkerboard to make it clear it's transparent.

PR #603, opened 2023-04-13
Disable deploy job in forks

Goooler:
See the failure in https://github.com/Goooler/Confetti/actions/runs/4689932575/jobs/8312328899.

### DIFF
--- a/.github/workflows/backend-deploy.yml
+++ b/.github/workflows/backend-deploy.yml
@@ -22,6 +22,7 @@ concurrency:
 jobs:
   deploy:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


martinbonnin:
We could also use the absence of secrets to disable the job. That'd scale a bit better if we ever decide to make this repo a template. No strong preference though, Happy to merge as is

martinbonnin:
> Something like if: ${{ secrets.GOOGLE_SERVICES_JSON != null }}?

I guess so (not sure about the syntax). Or maybe even do it in Gradle scripts? If no env variable, just do not deploy? Not 100% sure where I'm going there so let's merge like this and we can revisit when (if) we actually make this a template.

PR #610, opened 2023-04-15
Improve Wear now() logic

yschimke:


yschimke:
Going to land, but follow up reviews are welcome.

PR #611, opened 2023-04-16
Improve the bookmark screen by grouping clashing talks together.

brunohensel:
If the user bookmarks more than one talk for the same time slot, this clashing is not visible on the Bookmark screen. The start time of a talk is also not displayed on this screen.

This PR improves the user experience by introducing a stick header with the date and time information and grouping all talks for the same slot together on the Android project.


androidApp/src/main/java/dev/johnoreilly/confetti/bookmarks/BookmarksView.kt
```diff
@@ -156,17 +161,28 @@ private fun BookmarksHorizontalPager(
...
-                    conference = conference,
-                    session = session,
-                    sessionSelected = navigateToSession,
-                    isBookmarked = bookmarks.contains(session.id),
-                    addBookmark = { sessionId -> addBookmark(sessionId) },
-                    removeBookmark = { sessionId -> removeBookmark(sessionId) },
-                    onNavigateToSignIn = onSignIn,
-                    user = user,
-                )
+            val sessionsByDate = displayedSessions.groupBy { it.startsAt }
```
joreilly, line 164: Can be something we do in follow up PR if needed but wonder if we should do this in the view model....that way iOS client for example can also use data directly then (when we add bookmarks to it that is!)

brunohensel: Yes, I think it would be better to place this in the VM, and from there the consumers will receive the data already grouped, regardless of where the consumer is (which platform).

brunohensel: Does it make sense to have past sessions also grouped by date?

brunohensel: @joreilly  I decided to already provide the grouped type from the ViewModel as you suggested.

PR #614, opened 2023-04-16
AGP 8.0.0

joreilly:
Main changes affecting us

- namespace required in each module
- R classes are now non-transitive
- buildConfig now false by default


joreilly:
@yschimke just want to make sure this shouldn't cause you any issues on wear os side of things?

yschimke:
I doubt it, if so, let's find out. Submit, I'll check.

PR #629, opened 2023-04-20
Initial stab at french localization

martinbonnin:
I might have missed a few hardcoded strings there and there, will fix them along the way.


BoD (commented):
Sorry I know it's already merged, I used the GitHub mobile app which adds comments as pending, which I didn't  notice until now.

PR #630, opened 2023-04-20
Add adaptive monochrome icon

xxfast:
Took some creative liberty to add a monochrome version of the icon. I don't have a licence to sketch, so I can't open or modify the sketch file. I do have a figma source available.

Screenshot 2023-04-20 at 12 53 53 pm

This is how it looks 

Before:

After:

### DIFF
--- a/shared/src/androidMain/res/drawable/ic_launcher_foreground.xml
+++ b/shared/src/androidMain/res/drawable/ic_launcher_foreground.xml
@@ -1,0 +1,86 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="108dp"
+        android:height="108dp"
+        android:viewportWidth="449"
+        android:viewportHeight="448">
+    <group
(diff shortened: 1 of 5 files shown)


martinbonnin (approved):
Very cool, thanks! I'm guessing you took the SVG from the `.idea` folder as a base? 
Would be interesting to explore other design tools than sketch. Could we use Figma instead? Or does that require a license too? 

FWIW, plain SVG export from sketch loses some gradient/masks which is why I ended up commiting the sketch files but I'm not particularly happy about this situation

xxfast:
Thanks :) I ended up importing the `.sketch` file from figma, and it didn't seem to loose any gradient/masking. Here's the `.svg` for the monochrome version, and I will need help with someone with a sketch licence to commit this to source control for future reference 
C-bw

martinbonnin:
Nice! Today I learnt Figma can open .sketch files 👍

PR #634, opened 2023-04-20
Added Android Auto support to Confetti 🏎 🎉

cmota:
Hello John! 👋

I'm going to speak at Android Makers about Android Auto and I always like to create a sample so attendees can have a codebase to explore and follow along. Well, why not contribute to an already excellent project like Confetti 🙂.

I tried to use the same code conventions that are already in the project, and created the Android Auto project inside **androidApp/auto**.  I've also added a function to return the venue just so I can have an option to well, navigate directly to The Belfry of Montrouge.

To test the app you just need to install the DHU (Desktop Head Unit). You can find the tutorial here:
- https://developer.android.com/training/cars/testing

Ah! In case the app doesn't show up in the emulator, you need to go to the Android Auto app and select:
- Developer Settings > Unknown sources

Looking forward to your feedback! And of course, see you in one week 🚀


joreilly:
"I've also added a function to return the venue just so I can have an option to well, navigate directly to The Belfry of Montrouge."

@martinbonnin I think you added location info for venue to backend....not sure if it's exposed to client though...

joreilly:
ah, see that `venues` info is being retrieved from backend here...

martinbonnin:
Very very cool! Now we "just" need podcasting functionality to be able to listen to the talks in ~the traffic jams~ road trips :)

joreilly:
@cmota looks like we're getting following error after merging this (it's for release build which I guess we're not doing for PRs) 

```
> Task :androidApp:minifyReleaseWithR8 FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':androidApp:minifyReleaseWithR8'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.R8Task$R8Runnable
   > Compilation failed to complete

```

joreilly:
@cmota somewhat related to the above....we publish apps to play store....I'm not sure if it will see anything different with this  now that it supports Android Auto but we can add that as a form factor as well.

cmota:
@joreilly forgot to mention that you'll need to add Android Auto support on the Google Play Console, it's in the same place where you've added for Android Wear. Since I'm using templates it should be almost direct. Let me check if I can replicate in release to fix this.

cmota:
It seems I needed to update the proguard file. I've created a new PR to address this:
- https://github.com/joreilly/Confetti/pull/636

PR #635, opened 2023-04-21
FR Localization follow up

martinbonnin:
Feedbacks from https://github.com/joreilly/Confetti/pull/629#pullrequestreview-1394565511


BoD (approved):
👍

PR #642, opened 2023-04-21
compose title layout updates

joreilly:
Screenshot_20230421_222307


androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiScaffold.kt
```diff
@@ -143,8 +145,13 @@ fun ConfettiScaffold(
                 CenterAlignedTopAppBar(
                     title = {
                         if (title != null) {
-                            Text(text = title, fontSize = titleFontSize)
+                            Text(
```
yschimke, line 148: Is it worth setting maxLines = 2? 3?

yschimke (approved):

PR #643, opened 2023-04-22
Update proguard rules for Android Auto release and improved image fetcher for speakers avatars

cmota:
We need to add `-keep class` to `androidx.car.app.**` otherwise the app is unable to be launched in release mode.

Also improved the image fetcher for speakers' avatars, that sometimes wasn't loading the correct picture.


joreilly:
nice, I'll publish update to play store once this is merged

PR #646, opened 2023-04-22
Google Signin: Change the oauth client id

martinbonnin:
This somewhat makes Google signin work again. I'm not 100% sure what happened there.

The error that was happening was `ApiException: 12500`. From [stackoverflow](https://stackoverflow.com/questions/47632035/google-sign-in-error-12500), this can be a number of things from missing support email to wrong sha1

 I also put the Oauth consent screen back to "testing", which allows anyone to log in (with a warning)

From [doc](https://support.google.com/cloud/answer/10311615#publishing-status&zippy=%2Ctesting):

```
if your app requests a subset of the following: name, email address, and user profile (through the [userinfo.email, 
userinfo.profile, openid](https://developers.google.com/identity/protocols/oauth2/scopes#oauth2) scopes or their 
[OpenID Connect equivalents](https://developers.google.com/identity/protocols/oauth2/scopes#openid-connect)). 
For such requests, your users do not need to be in the trusted user list, they will not see a warning message, 
and their authorizations will not expire after 7 days.
```

This is our case, we request `userinfo.email` and `useringo.profile` only so I would say the "testing" mode allows is good enough for now? I'll try to hit the "publish" button once again after Android Makers see what this makes

### DIFF
--- a/androidApp/google-services.json
+++ b/androidApp/google-services.json
@@ -7,14 +7,22 @@
   "client": [
     {
       "client_info": {
-        "mobilesdk_app_id": "1:977739339802:android:0683e4e695beadd4b799a3",
+        "mobilesdk_app_id": "1:977739339802:android:bd3d868c82b383ccb799a3",
         "android_client_info": {
           "package_name": "dev.johnoreilly.confetti"
         }
       },
       "oauth_client": [
         {
-          "client_id": "977739339802-rtlrklolvmvd87ldes0qdc48ifcn30rh.apps.googleusercontent.com",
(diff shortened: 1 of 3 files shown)


martinbonnin:
For future reference, the current configuration
Screenshot 2023-04-22 at 18 13 37

Screenshot 2023-04-22 at 18 14 37

Screenshot 2023-04-22 at 18 14 51

Screenshot 2023-04-22 at 18 15 02

martinbonnin:
Turns out there is another problem R8 related. I think some backend configuration tweaks were needed but now I'm doubting everything... In all cases, better use the latest oauth credentials. Merging this one and will open a separate one for R8

PR #648, opened 2023-04-22
Reuse the Coil image cache for Auto

yschimke:
- Reuse the coil image cache
- Avoid side effects writing to image Map, and use updates to Flow instead.

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/auto/speakers/SpeakersScreen.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/auto/speakers/SpeakersScreen.kt
@@ -36,7 +36,7 @@ class SpeakersScreen(
     speakersViewModel: SpeakersViewModel = SpeakersViewModel()
 ) : Screen(carContext) {
 
-    val fallbackImage = CarIcon.Builder(
+    private val fallbackImage = CarIcon.Builder(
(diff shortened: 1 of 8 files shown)


yschimke:
cc @cmota - I'm not familiar with Auto, but wanted to reuse the image cache that Coil has, and use a similar approach for flows of data.

androidApp/src/main/java/dev/johnoreilly/confetti/auto/speakers/SpeakersScreen.kt
```diff
@@ -15,86 +17,117 @@ import dev.johnoreilly.confetti.SpeakersUiState
...
 class SpeakersScreen(
     carContext: CarContext,
     conference: String,
-    private val speakersViewModel: SpeakersViewModel = SpeakersViewModel()
+    speakersViewModel: SpeakersViewModel = SpeakersViewModel()
 ) : Screen(carContext) {
 
-    private var images: MutableList<AutoImage> = mutableListOf()
-    private var uiState: SpeakersUiState = SpeakersUiState.Loading
+    val fallbackImage = CarIcon.Builder(
```
cmota, line 39: Small thing here but:
- `fallbackImage`, `speakersFlow` and `imagesFlow` could be set as private

PR #684, opened 2023-05-05
Sign out: also sign out from Google account

martinbonnin:
Signing out from Firebase does not sign out of Google so next time you would try to sign in you wouldn't be able to change Google accounts if needed


martinbonnin:
I also increased the Gradle memory to 4g as I hit a OOM [there](https://github.com/joreilly/Confetti/actions/runs/4897690427/jobs/8746020390?pr=684)

yschimke (approved):

PR #688, opened 2023-05-08
Use Decompose for shared navigation

arkivanov:
As discussed on Twitter, here is a draft just to demonstrate how it would look like. Only the main Android app is working - it looks and feels basically the same, just added some fancy screen transitions. Wear, Auto (commented out), and any other platforms or tests are not touched yet. This should be also compatible with SwiftUI if needed. Unit and integration testing should be also pretty easy.

Tests and checks should be failing for now.

Please let me know your thoughts, feel free to discard if not interested, or I can finish the rest.


arkivanov:
> is idea that Decompose would provide something similar instead?

Yes, there are two options. Either expose states via a `Flow` wrapper (like `CFlow`), or use `Value` provided by Decompose. IN the latter case, the SwiftUI side works as [explained in the docs](https://arkivanov.github.io/Decompose/component/overview/#swiftui-example).

joreilly:
I guess more general question then is whether view models have a place when using Decompose?  Are there ways of using Decompose to specifically address shared navigation alone or is expectation that those broader changes would be needed as well to do that?

arkivanov:
There is no such an entity as ViewModel in Decompose. In Jetpack Compose, the ViewModel plays two roles: 1. It retains in the back stack; 2. It retains over configuration changes.

With Decompose, the first case is handled automatically - components are retained in the back stack.

The second case can be [handled](https://arkivanov.github.io/Decompose/component/instance-retaining/) explicitly via `InstanceKeeper` if needed. However, in Confetti it is already handled by the app-scoped repository, so the app works just fine.

arkivanov:
I have updated the iOS app.

arkivanov:
Updated Android Auto app

joreilly:
Not sure if it needs a rebase or something but getting following for some reason 

```
gh pr checkout 688
From https://github.com/joreilly/Confetti
 ! [rejected]          refs/pull/688/head -> decompose  (non-fast-forward)
failed to run git: exit status 1
```

arkivanov:
Perhaps, that's because I force-pushed one of the commits

joreilly:
ah, maybe I have branch here from before....let me delete it and try again.....yep, that was it!

joreilly:
The failure before was related to jvm target and due to instance not running....should be addressed by https://github.com/joreilly/Confetti/pull/702

androidApp/src/main/AndroidManifest.xml
```diff
@@ -73,8 +73,8 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
-<!--
+
```
yschimke, line 76: Should these be enabled in main?

arkivanov: Seems like I forgot to comment those back after testing Auto.

androidApp/build.gradle.kts
```diff
@@ -155,9 +155,11 @@ dependencies {
     implementation(libs.compose.material)
     implementation(libs.compose.material.icons.core)
     implementation(libs.compose.material.icons.extended)
-    implementation(libs.compose.navigation)
     implementation(libs.coil.compose)
 
+    implementation(libs.decompose.decompose)
```
oianmol, line 160: do you need to `implement` this even if we have it specified it as `api` under shared's build.gradle.kts ?

`api(libs.decompose.decompose)` 🤔

arkivanov: Most likely we don't need! Thanks for spotting. Also we can look for unused dependencies in the shared module.

PR #647, opened 2023-04-22
Make Firebase work with AGP 8 and R8 full mode

martinbonnin:
Fixes Google SignIn
see https://github.com/firebase/firebase-android-sdk/issues/2124

PR #703, opened 2023-05-19
Fix proguard config for Wear

yschimke:
Follow up to https://github.com/joreilly/Confetti/pull/647#issuecomment-1555032551

### DIFF
--- a/wearApp/proguard-rules.pro
+++ b/wearApp/proguard-rules.pro
@@ -5,16 +5,8 @@
 -dontwarn kotlinx.serialization.KSerializer
 -dontwarn kotlinx.serialization.Serializable
 
-# Referenced by okhttp
--dontwarn org.bouncycastle.jsse.BCSSLParameters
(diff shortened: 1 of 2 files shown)


wearApp/proguard-rules.pro
```diff
@@ -5,16 +5,8 @@
 -dontwarn kotlinx.serialization.KSerializer
 -dontwarn kotlinx.serialization.Serializable
 
-# Referenced by okhttp
```
yschimke, line 8: This shouldn't be needed with OkHttp 5 (or 4.11)

Maybe for the shaded version?

martinbonnin: OkHttp was 4.9 when [this was introduced](https://github.com/joreilly/Confetti/commit/e057483e146518d98352209407333a52824b3c4b), let me try to remove this.

martinbonnin: Looks like that works well : https://github.com/joreilly/Confetti/pull/712

PR #770, opened 2023-06-19
Extract strings from HomeScreen

luizgrp:
### WHAT

Extract strings from `HomeScreen`.

### WHY

Enable localisation, as in https://github.com/joreilly/Confetti/pull/629


joreilly:
@luizgrp  just fyi that there's a `Confetti` channel in Kotlinlang slack if you want to chat about anything there as well

luizgrp:
@joreilly awesome, thanks! joined it :)

PR #787, opened 2023-06-30
[Wear] Use SectionedList in HomeScreen

luizgrp:
### WHAT

Use `SectionedList` in `HomeScreen`.

### WHY

- For the app: simplify build of screen with different sections;
- For the library: test API in a real app and look for improvements;


luizgrp:
Snapshot tests are failing, https://github.com/google/horologist/pull/1462 should address it. Will re-open once the fix is published and project is updated with the new release.

PR #795, opened 2023-07-04
Updated Decompose to 2.1.0-alpha-04 and enabled predictive back gesture

arkivanov:
Feel free to postpone the change if you don't like alpha versions.

https://github.com/joreilly/Confetti/assets/26204457/6e4593e9-9b9d-414f-9f7a-42aa53e6bfa5


joreilly:
Does this require Android 14 to work?

arkivanov:
> Does this require Android 14 to work?

It requires compileSdkVersion=34, and it works only on Android U. Older versions remain the same.

PR #798, opened 2023-07-06
Update Decompose to 2.1.0-alpha-05

arkivanov:
The new version fixes an issue: the predictive back animation is played when the back button is pressed on Android <U.


joreilly:
@arkivanov I just noticed this....do you want me to merge this or perhaps update to later version?

arkivanov:
It's ok to merge as it fixes a bug. For the latest version 2.1.0-alpha-06, it's better to also update [ObservableValue](https://github.com/joreilly/Confetti/blob/9a6d6393b01f958b993c156aa99dc87c6f06bfe1/iosApp/iosApp/ObservableValue.swift#L4) to the [latest code](https://github.com/arkivanov/Decompose/blob/master/sample/app-ios/app-ios/DecomposeHelpers/ObservableValue.swift) (it fixes possible memory leaks).

PR #824, opened 2023-08-01
refactor(ios) : adds consistent padding for the segmented control and fixes list scrolling beneath toolbar

carotkut94:
# What?

* The segmented control looks out of place because of the missing surrounding padding
* Fixes list scrolling beneath toolbar

# Result
|Before|After|
|-|-|
|image|image|

# List scrolling

|Before|After|
|-|-|
|||


carotkut94:
@joreilly pushed another commit to fix the list scrolling below toolbar

joreilly:
nice, thanks for making those.....the iOS client hasn't been getting as much attention as it deserves :)

PR #850, opened 2023-09-02
Wear Decompose

yschimke:
Wear support for Decompose.


yschimke:
FYI @arkivanov

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/decompose/ConferencesComponent.kt
```diff
@@ -26,7 +26,10 @@ interface ConferencesComponent {
     sealed interface UiState
     object Loading : UiState
     object Error : UiState
-    class Success(val conferenceListByYear: Map<Int, List<GetConferencesQuery.Conference>>) : UiState
+    class Success(val conferenceListByYear: Map<Int, List<GetConferencesQuery.Conference>>) : UiState {
+        val relevantConferences: List<GetConferencesQuery.Conference>
+            get() = conferenceListByYear.values.flatten()
```
arkivanov, line 31: Perhaps, better to make it `Lazy`.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
```diff
@@ -46,7 +50,10 @@ class MainActivity : ComponentActivity() {
                     viewModel.waitingOnThemeOrData
                 }
 
-                ConfettiApp(navController, viewModel)
+                ConfettiApp(
+                    component = appComponent,
+                    viewModel
```
arkivanov, line 55: Use name argument here as well?

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
```diff
@@ -57,13 +64,20 @@ class MainActivity : ComponentActivity() {
 
     override fun onNewIntent(intent: Intent?) {
         super.onNewIntent(intent)
-        navController.handleDeepLink(intent)
+
+        if (intent != null) {
+            val handled = appComponent.handleDeeplink(intent)
```
arkivanov, line 69: I think the deeplink may come in `onCreate` as well? Maybe pass the deeplink via `DefaultWearAppComponent` constructor? And here just call `setIntent` + `recreate`, to go through the constructor.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/bookmarks/BookmarksComponent.kt
```diff
@@ -0,0 +1,74 @@
...
+import dev.johnoreilly.confetti.wear.tile.TileUpdater
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.datetime.LocalDateTime
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+interface BookmarksComponent {
```
arkivanov, line 23: Why do we need separate components for Wear? Can we reuse the existing ones?

yschimke: I did except when the screen structure meant they were quite different.

I will review this one again and reuse if possible.

For some of the common ones I do need to follow up and make them less network aggressive in a per device way. So I want to share.

arkivanov: Yeah, maybe generalise some of the existing components, so that they could be reused for Wear. It should be fine to have some parts unused on Wear. Some things could be abstracted with an interface. But if the logic is really different, then it might be better to actually have them separate. So up to you.

I think you are lucky that you only replaced top components. Replacing an intermediate component (in the middle of the hierarchy) would be a bit tricker.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
yschimke: I don't think onCreate is called in this case, only onResume.  I'll change the onCreate constructor call, but think I still need this.

https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)

wearApp/src/main/java/dev/johnoreilly/confetti/wear/bookmarks/BookmarksComponent.kt
yschimke: I think a follow up. The current one has slight differences in screen structure. Would like to get this in first before refactoring mobile further.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
arkivanov: I think there are two scenarios.

1. The activity is already created - the activity instance is reused and the intent is delivered via `onNewIntent`.
2. The activity is not yet created (e.g. the process is killed) - the activity is created and the intent is delivered via `onCreate`.

What do you think?

wearApp/src/main/java/dev/johnoreilly/confetti/wear/bookmarks/BookmarksComponent.kt
arkivanov: Sounds good.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
yschimke: I seem to be getting a third.

The activity is already created - the activity instance is reused and the intent is delivered only via onCreate.

I've tried to follow this but it doesn't appear to work, I see the right intent, but I get the app state from before the deeplink.

```
    override val stack: Value<ChildStack<Config, Child>> =
        childStack(
            source = navigation,
            initialStack = { initialConfig(intent) },
            childFactory = this::buildChild,
        )
```

arkivanov: > the activity instance is reused and the intent is delivered only via onCreate

If onCreate is called then it's a new instance of activity. But regardless, there any saved state when this happens? E.g. the savedInstanceState argument is not null? What are the steps to reproduce?

yschimke: Actually it seemed to work.  My repro steps, some of which might show my unfamiliarity with Navigation.

1. Launch confetti, navigate to bookmarks, some non default screen.
2. Close the Wear emulator with 2nd button.
3a. Hit 2nd button again and launch Confetti from All apps view.  Should launch a fresh activity, back to Home screen.  But is showing the bookmarks.
3b. Hit the 1st button and launch Confetti from Recents.  Should reuse the existing activity, working.
3c. Launch from a complication, should launch a fresh activity, to the deeplink.
3d. Launch from a tile, should launch a fresh activity, to the deeplink.

**But don't spend time on it yet, I'll ask at work.**

wearApp/src/main/java/dev/johnoreilly/confetti/wear/decompose/SwipeToDismissBox.kt
```diff
@@ -0,0 +1,178 @@
...
+        get() = _scalingLazyColumnState.value
+
+    @Composable
+    fun createScalingLazyColumnState(
+        factory: ScalingLazyColumnState.Factory = ScalingLazyColumnDefaults.belowTimeText()
+    ): ScalingLazyColumnState {
+        val scalingLazyColumnState = factory.create()
+        _scalingLazyColumnState.value = scalingLazyColumnState
+
+        scalingLazyColumnState.state = rememberSaveable(saver = ScalingLazyListState.Saver) {
```
yschimke, line 70: @arkivanov I just added this so that the LazyColumn position is retained when you swipe back.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
arkivanov: Thanks! I've go some questions.

3a - why launching Confetti from All apps should go to the initial screen? Are there any guidelines?
3c and 3d - what does "to the deeplink" mean? Where the deeplink comes from?

yschimke: The details get into task affinities and task stacks. https://developer.android.com/guide/components/activities/tasks-and-back-stack and https://developer.android.com/guide/components/activities/recents. These are mobile concepts also.

But it's somewhat confusing on Wear even for users. There are different ways to launch apps from the SysUI.

1) The App Launcher, is the top level launches for your App.  you can have multiple e.g. Compose Email, vs see Inbox.

image

It's an alphabetical list of activities on Wear OS.  Samsung shows this as your app icons in a grid 
image

Clicking here should start the thing you clicked, even if it means using an existing activity and changing the navigation stack. 

 2) Recents. This shows what you are currently running.

image

Samsung uses a visual display 
image

Clicking here should take you back to where you were.

yschimke: For complications and tiles.  They should be contextual, so take you straight into a session, but whatever, it should be clear to the user what they will do.

So TaskStackBuilder is used with Androidx Navigation, to restore the synthetic backstack.
complication

tile

arkivanov: Thanks! This seems aligned with mobile phones. I have just played with this [guide](https://developer.android.com/training/app-links/deep-linking), as well as this [doc](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)).

In the manifest, without `android:launchMode="singleTop"`, there is a new instance of activity created for each deep link. Each activity is added to the task stack.

With `android:launchMode="singleTop"`, if there is no created activity, a new one is created (`onCreate` is called), and the deeplink is available via the `Activity#getIntent` method. If there is a created activity, then it gets activated and `onIntent` called with the deeplink in the method's `intent` argument.

In the second case, the ideal approach would be to do the following.

```kotlin
    override fun onNewIntent(intent: Intent) {
        super.onNewIntent(intent)
        setIntent(intent)
        recreate()
    }
```

This recreates the activity and starts it with the deeplink in the intent, as it has not been created before. So you can as usual read the deeplink via `Activity#getIntent` method from `onCreate`. However, this also preserves the activity state, and so the navigation stack is also preserved and the deeplink is not handled. I think currently we should not recreate the activity and just call the component for navigation with the deeplink from `onNewIntent`. I will think how this could be optimized in Decompose.

wearApp/src/main/java/dev/johnoreilly/confetti/wear/decompose/SwipeToDismissBox.kt
arkivanov: Interesting, I thought that the existing `dev.johnoreilly.confetti.wear.navigation.SwipeToDismissBox` (which you removed) should have been handling this already. E.g. it wraps child screens with `SaveableStateProvider` which should keep the state.

yschimke: It does work. The issue here is the app level scaffold needs access to the screen level scroll state.

Not sure how else to plumb it through.

No issue, just complication of wear with time text at top of the app

wearApp/src/main/java/dev/johnoreilly/confetti/wear/MainActivity.kt
yschimke: I'll leave as is, but add singleTop.

Thanks

arkivanov: Though, I have just tried deeplinks on a normal phone with `singleTop`. When there is an activity already created (e.g. launched from the home screen), and then I click on a deeplink in another app, then a new MainActivity is created and added  to the activity task of the app where I clicked the link. So there are two created activities now.

The behaviour I described in the previous message (with saved state preventing the deeplink from being opened) is reproducible with adb: `./adb shell am start -W -a android.intent.action.VIEW -d "<link>" <pkg>`. Or by using `singleTask` launch mode instead of `singleTop`.

yschimke: Gone with singleTask for now. I'll follow up at work and try to work out how to test for this behaviour.

PR #894, opened 2023-09-22
Fix google dependencies

yschimke:


joreilly:
@yschimke just fyi that there's conflict with this....

yschimke:
Should be good now

PR #906, opened 2023-09-27
add DevFest Georgia 2023

tatocaster:


martinbonnin:
I'll merge this and trigger an import. The conference will appear after that.

PR #943, opened 2023-10-20
Empty bookmarkView

percy-g2:
Screenshot_1697821761


joreilly:
I think we need to check `upcomingSessions` as well?

percy-g2:
yes added upcomingSessions check as well.

percy-g2:
@joreilly now showing empty view separately for past and upcoming page now, as it makes more sense to show with tabs present.

PR #946, opened 2023-10-22
Fix bookmark icon state change issue in session list view.

percy-g2:


percy-g2:
@joreilly i'm still facing few issues while bookmarking and it doesn't return proper data in Success data class having bookmarks sets of session ids, would it be better to add a callback or add async await then update that uiState ? which get's reflected ?

joreilly:
@martinbonnin is more familiar with bookmarks implementation.....he'd probably better person to comment on this.

percy-g2:
sure, awaiting his feedback for this! Else this fixes for session list view for time being.

martinbonnin:
@percy-g2 can you describe the issue you're seeing? Is it the same as this one from the Kotlin slack? https://kotlinlang.slack.com/archives/C051P2HUVKP/p1694715876671799?thread_ts=1694712937.764289&cid=C051P2HUVKP

If yes, there's a transient state when the network request fails. I've been wanting to remove the optimistic updates for a while now. Sounds like a good time to do it.

percy-g2:
@martinbonnin i don't have access to that slack to conform it.

martinbonnin:
You can get an invite [here](https://surveys.jetbrains.com/s3/kotlin-slack-sign-up)

Else, it's available read-only [here](https://slack-chats.kotlinlang.org/t/15709817/anyone-here-at-droidcon-ny-slightly-smiling-face#b1f4d9be-fa6b-4646-966f-ca0196142eb7)

percy-g2:
awaiting approval for my email Id , until then i'm marking this MR/PR as daft or stale.

PR #951, opened 2023-10-23
Make full view clickable of conference in iOS

percy-g2:
https://github.com/joreilly/Confetti/assets/4012000/cf10b9f0-e2f0-4e70-b854-76096137c993

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-12
+    runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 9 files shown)


percy-g2:
@joreilly  please review

percy-g2:
@joreilly any update on this?

joreilly:
What's the issue this is fixing?  I tried both here and couldn't see difference but might be missing something

Also, can we keep this PR to just the swift changes (I'd like if possible for `SessionItemView` to only know about information relating to itself e.g. not needing full set of bookmarks to be passed to it)

percy-g2:
> What's the issue this is fixing? I tried both here and couldn't see difference but might be missing something
> 
> Also, can we keep this PR to just the swift changes (I'd like if possible for `SessionItemView` to only know about information relating to itself e.g. not needing full set of bookmarks to be passed to it) (UPDATE: ah, I see you just did that :) )

if you click the white space in between name of conference and date, it's not clickable like we have in android the whole card view and it doesn't navigate. It fixes that.

joreilly:
ah, ok....thanks

PR #981, opened 2023-11-10
Use a responsive Wear layout

yschimke:
New devices on the market are at 233dp, already larger than the Large Round emulator.  So switch to responsive layout, instead of hardcoded for 192 & 227 dp.

image

image


joreilly:
Nice!

PR #1022, opened 2023-11-29
Update dependencies

yschimke:
Seems some missed by renovate.


yschimke:
Most of the image diffs are really trivial
image

percy-g2:
I did the same and it broke my git branch, in my opinion we should migrate one by one based on screenshot testing! ACK

yschimke:
@percy-g2 this seems to be working fine, so landing.  The screenshots all come from one particular dependency (horologist) which improved layout.

PR #1041, opened 2023-12-05
Image loading progress logic for shared UI.

percy-g2:


percy-g2:
@joreilly i have a question regarding tinting icon according to theme in shared UI, can we do that in single place in here or call the native and check then return it ? because person icon needs to be tinted according to device current theme.

joreilly:
@percy-g2 not sure tbh....not something I've looked in to yet.

PR #1060, opened 2023-12-18
Add Compose WebApp

martinbonnin:
Revive https://wasm.confetti-app.dev/. It's now using 100% published artifacts.

It's currently very small (only lists conferences) but could be handy to build an admin portal to add/edit conferences easier than what we have now.


BoD (approved):
👍

PR #1062, opened 2023-12-19
Switch to Horologist AppScaffold

yschimke:
A better reusable implementation of App wide scaffolding.


joreilly:
Nice, looks a lot cleaner

PR #1073, opened 2023-12-21
Simplify TimeText / Scaffold

yschimke:


yschimke:
Mopping up some layout bugs as well.

PR #1076, opened 2023-12-25
Implement image resizing in the server

yschimke:
First cut of image resizing

- An additional endpoint ("https://confetti-app.dev/images/avatar/${data.conference}/${data.speaker.id}/Watch") that resizes images on the fly, without caching.
- Wear app calls the new input via Coil indirection. SpeakerAvatar -> AvatarMapper -> String.

Alternatives

- change the graphql server to optionally send back alternative photoUrl immediately? Either send both photoUrl + photoUrlResized. Or based on some flag?


wearApp/src/main/java/dev/johnoreilly/confetti/wear/speakerdetails/AvatarMapper.kt
```diff
@@ -0,0 +1,16 @@
+package dev.johnoreilly.confetti.wear.speakerdetails
+
+import coil.map.Mapper
+import coil.request.Options
+
+class AvatarMapper: Mapper<SpeakerAvatar, String> {
+    override fun map(data: SpeakerAvatar, options: Options): String? {
+        // TODO change to check A/B config from firebase
```
yschimke, line 8: @joreilly I was considering launching behind an A/B experiment. But probably overkill.

martinbonnin (approved):

yschimke:
@martinbonnin I trimmed it down to just the fetching end point, passed through as an additional field in the GraphQL response.

yschimke:
Will land so I can test the Wear code against a real server.

Issue #1045, opened 2023-12-06
Resizing image service


yschimke:
The avatar images have been observed to be 800kb.  For Wear we should resize the images down to 2-3kb. Maybe with some solution that allows mobile to use better sized images also.

yschimke:
Discussion on https://slack-chats.kotlinlang.org/t/16064427/do-we-have-some-good-options-for-resizing-images-i-assume-th#3132dcb0-5d91-4874-a5d2-44e0704c7ba9

PR #1078, opened 2023-12-28
Fetch thumbnails on Wear

yschimke:
Switch to thumbnails which should be smaller on Wear.

#fixes https://github.com/joreilly/Confetti/issues/1045


yschimke:
@martinbonnin not working currently - https://confetti-app.dev/images/avatar/devfestsrilanka2023/7879f060-41de-4095-b7ff-007bac7f52be returns NoSuchKey.

martinbonnin:
Should be fixed with https://github.com/joreilly/Confetti/pull/1082.

It's still failing on this though:

```
java.util.NoSuchElementException: null
	at kotlinx.coroutines.reactor.MonoKt.awaitSingle(Mono.kt:81)
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
```

yschimke:
@martinbonnin ready for review

martinbonnin:
Looks great 👍 I had mostly feedbacks about the backend side of things ([this PR](https://github.com/joreilly/Confetti/pull/1076)):

* What about adding arguments to `photoUrl`:

```graphql
type Speaker {
  """
  Url to the speaker photo or null if the speaker doesn't have a photo
  """
  photoUrl(
    """
    Maximum height of the photo. If the original photo has a height greater than maxHeight,
    it will be resized to maxHeight, preserving aspect ratio.
    If null, the photo is not resized.
    """
    maxHeight: Int = null
  ): String
}
```

This way we don't need to add a duplicate name which might be unclear (what is the difference between `photoUrl` and `photoUrlThumbnail`?) and it explicits `maxHeight`. 
Also, if we wanted, we could go all in with [`contentScale`](https://developer.android.com/jetpack/compose/graphics/images/customize#content-scale) later on.

* One inconvenient with `images/avatar/${conf}/${id}` is that it makes cache invalidation harder. If the photo content changes the url stays the same and the client could keep a photo around longer than needed (or we lower the TTL but that has a cost). Since "url" is opaque to the client, we don't have to do this now but what about url-encoding the original url? Or maintaining a hashtable `hash -> original url` somewhere maybe?

yschimke:
OK, I will take a look. That makes sense.

I didn't want to deal with all the cache headers in the server, but that's the real answer.  propogate ETags etc...

martinbonnin:
> I didn't want to deal with all the cache headers in the server, but that's the real answer. propogate ETags etc...

Not overly familiar with `ETags` so I was thinking of "just" setting a very long `maxAge` and have some sort of "content-addressable-url" to notify the client of changes. Since it goes to the GraphQL API first, that would trigger the fetch of the "new" content. But `ETags` sounds even more flexible. In all cases, this doesn't have to be done now. This can be left for "future optimisation" :)

yschimke:
OK, I'll land now.  I'm going to rely on GraphQL not failing if we ever remove/rename the additional field and just returning null.

yschimke:
FWIW I never trusted the caching headers on Wear for performance reasons, it sets ignoreCacheHeaders and caches until the URL changes.

PR #1084, opened 2023-12-29
Fix iOS crash - Add missing dependency definition inside KoiniOS

brunohensel:
The App crashes on iOS due to a missing dependency definition inside the `KoiniOS` file.

When we try to navigate to session details, the app crashes with the exception:

```
org.koin.core.error.NoBeanDefFoundException: No definition found for type 'com.apollographql.apollo3.cache.normalized.FetchPolicy'. Check your Modules configuration and add missing type and/or qualifier!
```

https://github.com/joreilly/Confetti/assets/45492687/5aa23f57-b41d-49e7-b4c0-3d157796bc9b

The navigation flow is fixed after adding the missing dependency definition. 

https://github.com/joreilly/Confetti/assets/45492687/52fd0757-1c0e-4759-b24d-7b6a4515693c

### DIFF
--- a/shared/src/iosMain/kotlin/dev/johnoreilly/confetti/di/KoiniOS.kt
+++ b/shared/src/iosMain/kotlin/dev/johnoreilly/confetti/di/KoiniOS.kt
@@ -3,6 +3,7 @@
 package dev.johnoreilly.confetti.di
 
 import com.apollographql.apollo3.ApolloClient
+import com.apollographql.apollo3.cache.normalized.FetchPolicy
 import com.apollographql.apollo3.cache.normalized.api.NormalizedCacheFactory
 import com.apollographql.apollo3.cache.normalized.sql.SqlNormalizedCacheFactory
 import com.russhwolf.settings.ExperimentalSettingsApi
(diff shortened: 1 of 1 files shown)


joreilly:
Thanks for catching this.....looks like it crept in as part of following commit https://github.com/joreilly/Confetti/commit/4ed78e7520c186e29f8541fcb122cf4cc09f1b3d ....hopefully we can add some kind of test to verify koin graph on iOS to catch regressions like this

joreilly:
it's strange....getting build error in Xcode now with those changes....if I revert locally then it builds/runs fine again.

```

Showing All Messages
> Task :shared:linkDebugFrameworkIosSimulatorArm64 FAILED



 * Source files: 

 * Compiler version: 2.0.0-Beta1

 * Output kind: FRAMEWORK



Compilation failed: Assertion failed

	at org.jetbrains.kotlin.backend.konan.llvm.DebugUtilsKt.location(DebugUtils.kt:269)

	at org.jetbrains.kotlin.backend.konan.llvm.DebugUtilsKt.lineAndColumn(DebugUtils.kt:275)

	at org.jetbrains.kotlin.backend.konan.llvm.CodeGeneratorVisitor.startLineAndColumn(IrToBitcode.kt:2284)

	at org.jetbrains.kotlin.backend.konan.llvm.CodeGeneratorVisitor.location(IrToBitcode.kt:919)

	at org.jetbrains.kotlin.backend.konan.llvm.CodeGeneratorVisitor.visitFunction(IrToBitcode.kt:881)

```

joreilly:
ok, looks like this was caused by ksp version update....have PR open to revert that

PR #1164, opened 2024-03-02
Use Roborazzi for Wear

yschimke:
Switch to roborazzi for Wear.

For follow up, I'll change the mobile as well. This seems to have one the RNG screenshotting space.

cc @oas004


yschimke:
The time text will be fixed in combination with https://github.com/joreilly/Confetti/pull/1163

But I'd like to land changes separately.

PR #1168, opened 2024-03-02
Wear: Use specific content layout 

yschimke:
Uses the responsive components.


yschimke:
@joreilly requesting post review, would like to make some follow on updates.

PR #1176, opened 2024-03-08
Wear conference colors

yschimke:
This probably suggests a cleanup is needed. Two reasons

- Currently conference is driven mainly by mobile if set.  So it means we need to read the DataLayer, then read Apollo to get the conference color.  
- The navigation theoretically allows for browsing other conferences with deeplinks, but this wont use the correct color


wearApp/snapshot/ConferenceHomeScreenTest/conferenceHomeScreenWithBookmarks[GenericLargeRound]_end.png
Comment (yschimke) [inline wearApp/snapshot/ConferenceHomeScreenTest/conferenceHomeScreenWithBookmarks[GenericLargeRound]_end.png:1]: @joreilly I'm not super happy with the Chip secondary colors.  But will dig in before rushing a fix.

joreilly:
looks great!

joreilly (approved):

PR #1179, opened 2024-03-08
theme cleanup + dark mode fix

joreilly:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/ui/ConferenceMaterialTheme.kt
```diff
@@ -0,0 +1,55 @@
...
+    val darkThemeConfigString by appSettings.settings.getStringFlow("darkThemeConfigKey", DarkThemeConfig.FOLLOW_SYSTEM.toString())
+        .collectAsState("FOLLOW_SYSTEM")
+    val darkThemeConfig = DarkThemeConfig.valueOf(darkThemeConfigString)
+    val shouldUseDarkTheme = shouldUseDarkTheme(darkThemeConfig)
+
+
+    var seedColor = Color(0xFF008000) // default if none set
+    seedColorString?.let {
+        try {
+            seedColor = Color(seedColorString.hexToLong(HexFormat { number.prefix = "0x" }))
```
joreilly, line 33: @yschimke just fyi that I had to switch to this in shared code as other java Long parsing doesn't work here.....maybe cleaner way still to do it

PR #1185, opened 2024-03-10
Improve conference switching for Wear

yschimke:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/decompose/AppComponent.kt
```diff
@@ -59,6 +59,9 @@ class DefaultAppComponent(
             } else {
                 val conferenceThemeColor = repository.getConferenceThemeColor()
                 showConference(conference = conference, conferenceThemeColor = conferenceThemeColor)
+
+                // Take the opportunity to update any listeners of the conference
```
yschimke, line 63: Extra work on every startup

wearApp/src/main/java/dev/johnoreilly/confetti/wear/navigation/WearAppComponent.kt
```diff
@@ -80,25 +81,16 @@ class DefaultWearAppComponent(
     private val networkRepository: NetworkRepository by inject()
     private val dataRequestRepository: DataRequestRepository by inject()
 
-    val conferenceDataFlow = phoneSettingsSync.conferenceFlow.map {
-        if (it.isNotBlank()) {
-            repository.conferenceData(it, FetchPolicy.CacheFirst).data
```
yschimke, line 85: This is really the point of this, want to be able to avoid an Apollo call before knowing where to do.

joreilly (approved):

yschimke:
@joreilly one nice thing. That test had caught a bug I chose to ignore. Now reenabled!

PR #1218, opened 2024-03-29
Fix serving JSON files with GCP

martinbonnin:


StylianosGakis (approved):
Exciting! Let's hope this gets us over the finish line!

PR #1222, opened 2024-03-29
Conference deep link for iOS

StylianosGakis:
For iOS:
I got into the docs here https://developer.apple.com/documentation/xcode/supporting-associated-domains#Add-the-associated-domain-file-to-your-website

And looked into how this looks like from firebase for the Hedvig app https://hedvig.page.link/.well-known/apple-app-site-association

Then the article here mentions that the identified before the app’s package name is the teamID. So for “appID” it should be “$teamID.$bundleID”.
Not 100% confident this is correct, but when I opened the Xcode project, for the team it said “Unknown name (NT77748GS8)” so I assume either “NT77748GS8" is the right name, or something that only exists on my machine for some reason. This we need to ensure before moving forward. From some forum they say that it should be visible under https://developer.apple.com/account/#/membership and then “This will get you to your Membership Details, just scroll down to Team ID”, so I guess John is the only one who can see this to confirm.


StylianosGakis:
https://github.com/joreilly/Confetti/commit/d67084ac99c839bf733d1ec519423069a193c7d6 seems to have worked.
Video recording below:

https://github.com/joreilly/Confetti/assets/44558292/3056e66f-c012-4f3a-a0bf-b2154ea25405

Copy paste this to test yourself too if you want
`xcrun simctl openurl booted "https://confetti-app.dev/conference/kotlinconf2023"`
Change the conference name to whatever you want it to be.
Also tried with `kotlinconf2023` which does not exist yet to show the error scenario too

StylianosGakis:
Current behavior of the deep links, with Decompose 3.x.
I do not know if there's something equivalent to "process death" I can test for here? :D 

Turning this back into a draft so that we can handle two things better:
* Destroying the old component's lifecycle when we are no longer using it.
* Figuring out a way to make AppComponent properly observable to remove the `reRenderIndex` hack

StylianosGakis:
The current behavior after commit c79600a8f2d0f992d2756bcca8466909c5a8b9a0 with observable object and all working
https://github.com/joreilly/Confetti/assets/44558292/ebba637a-38e8-4388-bba7-ce20525f90ce

iosApp/iosApp/AppDelegate.swift
```diff
@@ -11,17 +11,41 @@ import ConfettiKit
 
 class AppDelegate : NSObject, UIApplicationDelegate {
     
-    let root: AppComponent
```
arkivanov, line 14: Maybe we could conform `AppDelegate` to `ObservableObject` for simplicity? And just make this `root` property `@Published var`. Can you try that? If for some reason this is not possible or not working, then maybe move the `AppComponent` creation and replacing logic to `AppComponentHolder`?

Otherwise looks good!

StylianosGakis: Hmm yeah I didn't try to make `AppDelegate` an `ObservableObject` since I wasn't sure if that would be a weird thing to be that and an `UIApplicationDelegate` at the same time. I will give that a shot.

If that does not work for whatever reason, when you say "maybe move the AppComponent creation and replacing logic to AppComponentHolder" I assume you mean so that `AppDelegate` just calls a small function in `init()` and on `onConferenceDeepLink` so that the bulk of the work is done in there instead.

Will pick this up some next week, hopefully earlier rather than later inside of it. Thanks for the feedback 😊

arkivanov: >  when you say "maybe move the AppComponent creation and replacing logic to AppComponentHolder"

Yes. Just create the `AppComponent` together with its `ComponentContext` and `Lifecycle` inside `AppComponentHolder`. And also add `onConferenceDeepLink` to `AppComponentHolder` instead, and call that from `AppDelegate`. I think this would be cleaner. But the first option with conforming `AppDelegate` to `ObservableObject` sounds even better.

arkivanov (approved):
Thanks!

iosApp/iosApp/AppDelegate.swift
StylianosGakis: Hey making AppDelegate conform to ObservableObject just worked, thanks for your tips!

PR #1238, opened 2024-04-05
initial changes for service type...and avoiding showing as lightning talk

joreilly:


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/Model.kt
```diff
@@ -6,8 +6,10 @@ import kotlinx.datetime.TimeZone
 import kotlinx.datetime.toInstant
 
 fun SessionDetails.isBreak() = this.type == "break"
+fun SessionDetails.isService() = this.type == "service"
```
martinbonnin, line 9: Will add an enum for that so we have better type safety and we can map various conferences to that enum

martinbonnin (approved):

PR #1275, opened 2024-04-17
OkHttp version bump

yschimke:


yschimke:
Blocked on https://github.com/square/okhttp/issues/8369

PR #1279, opened 2024-04-19
Use roborazzi for screenshots

yschimke:
Drop quickbird, use roborazzi.

This won't actually run on CI automatically, I will follow up with the workflow changes to activate.


yschimke:
cc @oas004 - didn't see much uptake so I didn't keep the A11y test.  It would be trivial to do so.

yschimke:
Test only

oas004:
> cc @oas004 - didn't see much uptake so I didn't keep the A11y test. It would be trivial to do so.

I think that is fine, sorry for not being that active in this lately. Have a bit much on my plate at work lately :(

yschimke:
@oas004 I have zero expectations of your time, you never need to apologise.

PR #1292, opened 2024-04-20
Remove unexpected quote escaping in string resource

SimonMarquis:
Screenshot_20240420-154012

### DIFF
--- a/shared/src/commonMain/composeResources/values/strings.xml
+++ b/shared/src/commonMain/composeResources/values/strings.xml
@@ -31,7 +31,7 @@
     <string name="oops">Oops something went wrong</string>
     <string name="retry">Retry</string>
     <string name="update_wear">Update Wear Theme</string>
-    <string name="sign_in_for_bookmarks">In order to save your bookmarks, you\'ll need to sign in.</string>
+    <string name="sign_in_for_bookmarks">In order to save your bookmarks, you'll need to sign in.</string>
     <string name="cancel">CANCEL</string>
     <string name="sign_in">SIGN IN</string>
     <string name="sign_in_google">Sign in with Google</string>
--- a/androidApp/src/main/res/values/strings.xml
+++ b/androidApp/src/main/res/values/strings.xml
@@ -31,7 +31,7 @@
     <string name="oops">Oops something went wrong</string>
     <string name="retry">Retry</string>
     <string name="update_wear">Update Wear Theme</string>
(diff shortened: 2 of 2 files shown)


martinbonnin (approved):
👍

SimonMarquis:
Oh, interesting.

```
/home/runner/work/Confetti/Confetti/androidApp/src/main/res/values/strings.xml:36:4: Failed to flatten XML for resource 'sign_in_for_bookmarks' with error: Invalid unicode escape sequence in string
"{str}"
```

Not sure what's wrong. Wrapping in double quotes would probably fix the issue. But testing this is kinda tricky on a smartphone 😅

shared/src/commonMain/composeResources/values/strings.xml
```diff
@@ -31,7 +31,7 @@
     <string name="oops">Oops something went wrong</string>
     <string name="retry">Retry</string>
     <string name="update_wear">Update Wear Theme</string>
-    <string name="sign_in_for_bookmarks">In order to save your bookmarks, you\'ll need to sign in.</string>
+    <string name="sign_in_for_bookmarks">In order to save your bookmarks, you'll need to sign in.</string>
```
martinbonnin, line 34: Looks like this one works without the `expect` but the other one needs it...

SimonMarquis:
@martinbonnin @joreilly here are some related issues:
- https://github.com/JetBrains/compose-multiplatform/issues/4178
- https://github.com/JetBrains/compose-multiplatform/issues/4664

that brings up the question of why are we trying to mirror string resources in `androidApp/src/main/res/values/strings.xml` since they don't override the ones defined in `shared/src/commonMain/composeResources/values/strings.xml`?

It seems like resources that are directly used in the shared module can't be modified by regular Android string resources. What do you think about declaring those only in the shared module (with the `-fr` alternatives)?

```xml
<resources>
    <string name="sessions">Sessions</string>
    <string name="oops">Oops something went wrong</string>
    <string name="retry">Retry</string>
    <string name="sign_in_for_bookmarks">In order to save your bookmarks, you'll need to sign in.</string>
    <string name="cancel">CANCEL</string>
    <string name="sign_in">SIGN IN</string>
</resources>
```

joreilly:
Hmm, I think I might have been in process of transitioning those strings to shared code which is why they're probably still duplicated.  Will take a look soon at cleaning that up.

PR #1363, opened 2024-06-04
German translation improvements

rschattauer:


shared/src/commonMain/composeResources/values-ie/strings.xml
```diff
@@ -46,4 +46,6 @@
     <string name="empty_text">Gan aon data i láthair</string>
     <string name="no_bookmarks">Gan aon leabharmharcanna leagtha</string>
     <string name="privacy_policy">Beartas Príobháideachta</string>
+    <string name="bookmarks_past">Caite</string>
+    <string name="bookmarks_upcoming">Le teacht</string>
```
rschattauer, line 50: no idea if these are correct, feel free to change them

shared/src/commonMain/composeResources/values-fr/strings.xml
```diff
@@ -44,4 +44,6 @@
     <string name="no_paired_watch">Pas de montre appairée</string>
     <string name="no_bookmarks">Aucun favori défini</string>
     <string name="privacy_policy">Politique de confidentialité</string>
+    <string name="bookmarks_past">Passé</string>
+    <string name="bookmarks_upcoming">À venir</string>
```
rschattauer, line 48: no idea if these are correct, feel free to change them

androidApp/src/main/java/dev/johnoreilly/confetti/bookmarks/BookmarksView.kt
```diff
@@ -92,9 +94,9 @@ private fun BookmarksContent(
     }
 }
 
-private enum class BookmarksTab(val title: String) {
-    Past(title = "Past"),
-    Upcoming(title = "Upcoming")
+private enum class BookmarksTab(val title: @Composable () -> String) {
```
rschattauer, line 97: Either like this or returning StringResource type

joreilly:
Thanks for that!

shared/src/commonMain/composeResources/values-fr/strings.xml
joreilly: @martinbonnin wdyt? We can follow up with another PR if they do need to be changed

shared/src/commonMain/composeResources/values-de/strings.xml
```diff
@@ -1,18 +1,18 @@
 <resources>
     <string name="schedule">Zeitplan</string>
-    <string name="sessions">Sitzungen</string>
-    <string name="speakers">Redner</string>
+    <string name="sessions">Vorträge</string>
+    <string name="speakers">Sprecher</string>
     <string name="venue">Veranstaltungsort</string>
     <string name="search">Suchen</string>
     <string name="bookmarks">Lesezeichen</string>
-    <string name="rooms">Zimmer</string>
```
ashdavies, line 8: 😂

PR #1454, opened 2024-10-06
Caching dependencies for ios build action workflow

mirzemehdi:
for other environments I don't think caching makes that much difference, but for iOS, saw that it takes around 20 minutes for each build, this should make it 2 times faster for next builds.

Wasn't make sure if it uses CocoaPods, so added caching option for that too just in case.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -18,6 +18,33 @@ jobs:
           distribution: 'zulu'
           java-version: 17
 
+      - name: Cache KMP libraries
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.konan
+          key: ${{ runner.os }}-v1-${{ hashFiles('**/libs.versions.toml') }}
+
+      - name: Cache SPM downloads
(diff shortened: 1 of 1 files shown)


mirzemehdi:
@joreilly didn't know for pull request cache can work differently:

"When a cache is created by a workflow run triggered on a pull request, the cache is created for the merge ref (refs/pull/.../merge). Because of this, the cache will have a limited scope and can only be restored by re-runs of the pull request. It cannot be restored by the base branch or other pull requests targeting that base branch" (https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

For my case, workflow trigger was on pushing main branch, so I could see real benefit.

mirzemehdi:
There is also this: "If a workflow run is triggered for a pull request, it can also restore caches created in the base branch, including base branches of forked repositories.".

So It think best option would be just observing for next builds and see if it works as expected 😂

PR #1465, opened 2024-11-03
Remove excessive curly brackets on session tile deeplink

oas004:
In reference of https://kotlinlang.slack.com/archives/C051P2HUVKP/p1730368174142119.  Here after clicking the tile service we showed an empty screen instead of the session details

From what I could see we were sending excessive curly braces around the details in the deep link. This made the config values to store the ids with these curly braces.  I'm not sure if we actually need the curly braces here for some other reason. Maybe you know better @yschimke ? If that is the case we might want to consider stripping them from the build config path instead. For now I just removed them in the deep link.

After this change it works as expected 

https://github.com/user-attachments/assets/c59c0346-df80-44dc-8249-773e4e7179d1


yschimke (approved):

yschimke:
Nice fix, thanks

PR #1487, opened 2024-11-30
migrating to use CMP for all of UI

joreilly:
Still some work needed but pretty much all of the UI is now in shared CMP code
Screenshot 2024-11-30 at 18 33 38


joreilly:
@yschimke I need to follow up to add back settings option to connect android to wear device etc

androidApp/src/main/java/dev/johnoreilly/confetti/MainActivity.kt
```diff
@@ -45,27 +37,23 @@ class MainActivity : ComponentActivity() {
     @OptIn(ExperimentalDecomposeApi::class)
     override fun onCreate(savedInstanceState: Bundle?) {
         enableEdgeToEdge()
-        // Fix for three-button nav not properly going edge-to-edge.
-        // TODO: https://issuetracker.google.com/issues/298296168
-        window.setFlags(FLAG_LAYOUT_NO_LIMITS, FLAG_LAYOUT_NO_LIMITS)
```
Goooler, line 50: This shouldn't be removed due to #1395.

PR #1498, opened 2024-12-07
Attempt to use Credential Manager for sign-in on Wear

yschimke:
Refactor: Use Credential Manager for sign-in on Wear

note: this will only work on Wear 5.1, and isn't currently working, so just testing for the moment. I might provide support via simple token sharing otherwise.

Refactor the sign-in process to use the Credential Manager, replacing the previous Firebase sign-in screen. This change enables the app to utilize the system's credential store for a more streamlined user experience.

The `FirebaseSignInScreen` has been removed, and the sign-in functionality has been moved to a new `signIn` function within the `account` package. This function interacts with the Credential Manager to retrieve the user's Google ID token and subsequently signs them in using the `Authentication` service.

In the `MainActivity`, the `onSignIn` and `onSignOut` callbacks are now responsible for initiating the sign-in/sign-out flows, utilizing the `signIn` and `credentialManager.clearCredentialState` functions, respectively. This ensures that sign-in events are handled appropriately and user credentials are managed securely.


yschimke:
The token sharing from mobile appears to work, there is some more cleanup to come, but I'll land and then iterate on the different improvements.

PR #1519, opened 2024-12-15
dependency updates

joreilly:


joreilly:
@martinbonnin any concern with any of these versions?  I ran backend services locally and they at least built/run ok.

martinbonnin:
TBH I have no idea, let's merge and I'll monitor if it works 😄

martinbonnin:
All good ✅

PR #1537, opened 2025-02-12
Use Apollo incubating normalized cache

BoD:
Use Apollo's [incubating normalized cache](https://github.com/apollographql/apollo-kotlin-normalized-cache-incubating) instead of the main repository one.

Note: this version isn't "stable" yet (the API and storage can evolve in the upcoming versions, and while performance _should_ be improved this isn't completely confirmed) but I think Confetti is a good candidate for dogfooding it, e.g. it's ok for the cached data to be lost when upgrading - fresh data will be fetched from the backend.

Also in this PR I use the new [`@cacheControl`](https://apollographql.github.io/apollo-kotlin-normalized-cache-incubating/cache-control.html) directive with a max age of 4 hours (1 hour for bookmarks) - which is a bit arbitrary, LMK what you think about these values. ~What this means concretely is that cached data older than this will not be displayed and enforce a network call.~ The cached data older than this will still be displayed, but this will trigger a network fetch in the background.


joreilly:
>What this means concretely is that cached data older than this will not be displayed and enforce a network call.

Does this mean if you open app when offline then potentially that data won't show?

BoD:
In 92d6e11e0864a819746eccd025cb361e8a511630 I change this so it will display the cache data even if it's stale - but will also fetch fresh data from the network in the background when that happens.

joreilly (approved):

PR #1540, opened 2025-02-17
Tweak the custom fetch policy

BoD:
The interceptor introduced in #1537 fetches data in the background, which can make the UI change while the user is using the app without an action from their part. This can be confusing or annoying.

Instead, this version returns the data from the cache if it's fresh, if not, tries the network, and if that didn't work, falls back to the stale data from the cache.

Thanks @martinbonnin for the idea!


martinbonnin (approved):
No more rug pulling! 🎉

PR #1554, opened 2025-04-01
Add developer setting button to send notifications

yschimke:
Allow forcing sending notifications with a widened range for testing.


yschimke:
For follow up, we need to request permissions.

joreilly (approved):

PR #1595, opened 2025-04-09
Add Bookmarks Loading and Error states

yschimke:
This commit adds previews for loading and error states to the Bookmarks screen.

Those UIs don't currently exist.


yschimke:
cc @shumelchyk

PR #1701, opened 2026-07-20
ci(wear): enable the wear-scoped Compose Preview comment workflow

yschimke:
Turn compose-preview.yml from workflow_dispatch-only into an active per-PR visual-diff bot: render on push to main (baselines) and pull_request (diff + sticky comment). Two parallel jobs (compose / a11y), apply pinned to v0.16.57 with cli-version: catalog for CLI<->plugin lockstep (composeai-preview bumped to 0.16.57 to match).

Scope the render to :wearApp by dropping the composeai-preview plugin from :shared and :androidApp: those modules are minSdk 26 but pull ai.koog (40+ Android artifacts, all minSdkVersion 35), so the Robolectric preview render's debug unit-test manifest merge hard-fails before rendering. :wearApp renders clean; the phone modules were never diffed (the workflow was dispatch-only).

### DIFF
--- a/.github/workflows/compose-preview.yml
+++ b/.github/workflows/compose-preview.yml
@@ -1,9 +1,29 @@
 name: Compose Preview
 
-# Auto-triggers disabled until the GITHUB_TOKEN can push baseline/PR
-# branches (currently 403 against joreilly/Confetti.git). Run manually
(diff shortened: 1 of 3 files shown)


yschimke:
@copilot please fix the merge conflicts in this pull request.

PR #1750, opened 2026-07-29
fix(theme): complete the M3 colour schemes and make ConfettiTheme self-sufficient

yschimke:
Draft — three theme fixes found by sweeping the published `confetti-mobile` design catalog (https://preview.coo.ee/confetti-mobile/) in light and dark, then under all 14 declared `@ThemeCatalog` providers. Rebased onto `main` after #1751.

**Reworked in response to review.** The first version of this PR wrapped the catalog previews in a `Surface` inside `ComponentCatalog.kt`, which fairly read as hacking app code for previews. That is gone — `ComponentCatalog.kt` is now byte-identical to `main`, and the underlying gap is fixed in the theme itself instead. All three changes below are plain app fixes with no preview-specific code.

## 1. `ConfettiTheme` never set `LocalContentColor` (`Theme.kt`)

`MaterialTheme` installs the colour scheme but **not** the content colour — only a `Surface` does that. So any content hosted outside a `Surface` drew its text and icons in the CompositionLocal's default, `Color.Black`, whatever the scheme said. Nothing in the app hit it, because every screen sits under a `Scaffold`. But it left the theme unable to stand on its own: a caller that merely wraps content in `ConfettiTheme` got black text.

It is now anchored to `colorScheme.onBackground`. A `Surface` still overrides it with its own content colour, so nothing inside one changes.

This went from latent to visible the moment #1751 landed. `compose-preview` 0.19.6 carries [a fix](https://github.com/yschimke/compose-ai-tools/pull/2922) that paints a night `@Preview(showBackground = true)` on a dark backing instead of white — correct in itself, but it turned black-on-white into black-on-near-black:

content colour before and after

Top of each pair is `main` as it stands today; bottom is this PR. Measured: background `#1C1B1F`, text now `#ECDFE0` (`onBackground` in the dark scheme) where it was `#000000`. Light mode is unchanged — `#201A1B` on white, the light scheme's `onBackground`.

## 2. The colour schemes were missing the M3 surface ramp (`Theme.kt`, `Color.kt`)

`LightDefaultColorScheme`, `DarkDefaultColorScheme`, `LightAndroidColorScheme` and `DarkAndroidColorScheme` set `surface` and `surfaceVariant` but never the `surfaceContainer*` / `surfaceDim` / `surfaceBright` roles Material 3 added later. `lightColorScheme()` / `darkColorScheme()` default those to the **baseline (purple) palette** rather than deriving them from the `surface` you passed in — so the Android green scheme drew Cards, app bars, navigation bars and sheets on brand-purple neutrals.

Measured off the published catalog, the Android theme's ramp was byte-identical to the Confetti purple theme's:

| role | Confetti light | Android light (before) | Android light (after) |
|---|---|---|---|
| surfaceContainerLowest | `#FFFFFF` | `#FFFFFF` | `#FFFFFF` |
| surfaceContainerLow | `#F7F2FA` | `#F7F2FA` | `#F2F4EF` |
| surfaceContainer | `#F3EDF7` | `#F3EDF7` | `#EDEEEA` |
| surfaceContainerHigh | `#ECE6F0` | `#ECE6F0` | `#E7E9E4` |

Only `surface` itself (`#FBFDF7`) was actually green. The dynamic and conference-seeded schemes were already correct, because they are generated rather than hand-written.

surface ramp before and after

The new neutral tokens in `Color.kt` are generated rather than eyeballed: an M3 tone *is* the CIELAB `L*` value and a tonal palette holds hue and chroma constant across tones, so each token is its palette's mean `(a*, b*)` at `L* = tone`. The existing anchors already sit exactly on their tone's `L*` (`DarkPurpleGray10` measures `L*10.0`, `DarkPurpleGray90` measures `L*89.9`), so the ramp stays continuous. Tones follow the M3 mapping — light `N100/96/94/92/90` with dim `N87` and bright `N98`; dark `N4/10/12/17/22` with dim `N6` and bright `N24`.

## 3. One hardcoded colour (`SessionItemView.kt`)

The room name was `color = Color.Gray` — the only hardcoded colour left in `shared/…/ui`. Same `#888888` in both modes: ~3.5:1 on the light surface, below the 4.5:1 WCAG AA floor for body text, and unable to follow the theme at all. Now `MaterialTheme.colorScheme.onSurfaceVariant`.

## Android Studio

Nothing here is preview-harness-specific — `LocalContentColor` and the `surfaceContainer*` scheme parameters are ordinary Material 3 APIs, and no `@Preview` annotation or signature changes. Previews open and render in Studio exactly as before, and dark-mode ones get the same legibility fix for the same reason.

## Test plan

- `./gradlew :androidApp:composePreviewRenderAll` on 0.19.6 — rendered the full catalog; the images above are that output.
- Sampled pixels rather than eyeballing: dark previews are `#ECDFE0` on `#1C1B1F`; the light variants are unchanged at `#201A1B` on white.
- Sampled the surface-ramp swatches from the `Theme/Android` specimen to confirm the ramp is green-neutral rather than baseline purple.

## Please tell me if the evidence images should not be committed

`tmp-previews/theme-fix/*.png` are the before/after sheets above, committed only so they render inline here. The repo has no established home for PR evidence images and I did not want to invent one — happy to drop or move them.

## Follow-up (not in this PR)

The 14 theme chips on the published catalog are inert. The server-side half is fixed in yschimke/compose-ai-tools#2923 (merged, shipped in 0.19.6). The catalog-side half remains: `compose-preview` applies a `themeProvider` *in place of* a preview's `@PreviewWrapper`, but these previews call `CatalogTheme { … }` inside the function body, where an outer wrapper cannot reach. Migrating the catalog to declare its theme via `@PreviewWrapper` would make the chips live.


yschimke:
Triage of the preview-diff bot's report — three things worth reading before trusting it.

**The two render failures are pre-existing, not from this PR.** `androidApp/activity__MainActivity` and `wearApp/activity__MainActivity` produce no PNG on the base commit either:

```
androidApp/activity__MainActivity @ 25a54a0 (base) → 404
androidApp/activity__MainActivity @ 15ceed8 (head) → 404
wearApp/activity__MainActivity   @ 25a54a0 (base) → 404
wearApp/activity__MainActivity   @ 15ceed8 (head) → 404
```

**The Wear change (`HomeListViewKotlinConf`) is not reachable from this diff.** The head render is 112px taller than the base (1650 → 1762) and shows a **fourth day row, a duplicated "Friday"**, where `ConferenceFixtures.kotlinConf.days` declares exactly three (21–23 May 2025 = Wed/Thu/Fri). That looks like a slice-seam duplicate in the `@ScrollingPreview(modes = [ScrollMode.LONG])` stitched capture rather than a content change. Nothing in this PR can reach it: `wearApp` has its own `ui/Color.kt` with its own `internal DarkPurpleGray*` tokens and its own theme, and it does not use `SessionItemView` — the only `shared/` file this PR touches.

**The four `ThemeFoundation*Preview` entries are listed as "Unchanged", but that classification is not evidence — the head run published no PNG for them at all.** They exist on the base commit and are missing at head:

```
androidApp/ThemeFoundationAndroidPreview_Dark @ 25a54a0 (base) → 200
androidApp/ThemeFoundationAndroidPreview_Dark @ 15ceed8 (head) → 404
androidApp/ThemeFoundationDefaultPreview_Dark @ 25a54a0 (base) → 200
androidApp/ThemeFoundationDefaultPreview_Dark @ 15ceed8 (head) → 404
```

(same for `ThemeFoundationConferenceSeedPreview` and `ThemeFoundationDynamicPreview`). The bot's own "Render Failures" section counts only the two `MainActivity` captures, so these four fell into "Unchanged" by absence rather than by comparison. They render fine locally — `./gradlew :androidApp:composePreviewRenderAll` produced all eight PNGs — and the surface-ramp before/after in the PR description is that local output, not this CI run. So the ramp change is genuinely unverified by CI here; it is the one claim in the description that rests on a local render. Worth a look at why that run dropped them, though it is a harness question rather than something this PR changes.

---

PR #1753, opened 2026-07-31
fix(previews): let an overridden preview theme turn the app's own theme off

yschimke:
## The bug

The **Theme** selector on the published design catalogs does nothing. Every entry renders byte-identical pixels.

Reproducible against the live server right now:

```
GET /confetti-wear/render/sessioncard__ideal__default__compact.png
  ?themeProvider=…KotlinConfThemeCatalog        → md5 801c6414…
  ?themeProvider=…DevFestThemeCatalog           → md5 801c6414…
  ?themeProvider=…ConfettiDefaultThemeCatalog   → md5 801c6414…
  ?themeProvider=com.bogus.NotAThemeCatalog     → 400
```

Same on `confetti-mobile`, where a **light** and a **dark** provider also return identical bytes. It's not the toolchain: the `wear-m3` sample catalog on the same box *does* differ per theme, and the `400` proves the real providers resolve and get invoked.

These two URLs are live and currently render the same image, which is the bug — after this lands and the catalog regenerates they should differ:

| KotlinConf | DevFest |
|---|---|
| [](https://preview.coo.ee/confetti-wear/render/typography__ideal__default__compact.png?themeProvider=dev.johnoreilly.confetti.wear.ui.KotlinConfThemeCatalog) | [](https://preview.coo.ee/confetti-wear/render/typography__ideal__default__compact.png?themeProvider=dev.johnoreilly.confetti.wear.ui.DevFestThemeCatalog) |

That's the *Typography* specimen, which is supposed to show JetBrains Mono for KotlinConf and Google Sans Flex for DevFest.

## Cause

Every catalog `@Preview` installs the theme in its own body — `ConfettiThemeFixed { … }` on Wear, `CatalogTheme { … }` on the phone. That's deliberate and worth keeping: it's what makes the Android Studio preview pane and a plain `compose-preview` render show the real app theme with no annotation and no tooling in the loop.

But the preview server applies an alternative theme by *wrapping* the preview function, so the body's `MaterialTheme` composes **inside** that wrapper and shadows it. `ConfettiThemeFixed` is `MaterialTheme(ColorScheme(), Typography())` — hardcoded Wear defaults, reading nothing from outside — so whatever the daemon wrapped around it contributes exactly nothing.

## Fix

Keep the theme where it is and give it an off switch.

`LocalPreviewThemeOverride` (in `:shared`, so Wear, phone and the CMP targets share one definition) defaults to `false`. Android Studio, the CLI, and the app itself are untouched — the default theme is still the one the preview body installs.

The `@ThemeCatalog` / `@WearThemeCatalog` providers mark the theme they install with `PreviewThemeOverrideInstalled`, and `ConfettiThemeFixed` / `ConfettiTheme` / `ConferenceMaterialTheme` stand down when they see it rather than installing a second theme on top.

Standing down **consumes** the flag, so only the outermost app theme yields. Deliberately nested theming still applies — `ConferencesView` keeps tinting each row with that conference's own seed colour under an override instead of the override flattening it.

Two things fall out of putting this in the theme functions rather than on each preview: it covers previews an annotation-based fix would have missed (`AndroidScreenPreviews` among them), and `CatalogTheme`'s Coil preview handler survives an override for free — only the theme yields, not the rest of the preview environment.

**Tradeoff worth flagging:** the coupling is implicit. A new theme function that forgets the two-line check silently breaks the switcher again. There's a note in the `LocalPreviewThemeOverride` KDoc, but it's a convention, not a gate. The alternative — declaring the theme per-preview via androidx's `@PreviewWrapper` — is explicit and machine-readable, but needs an annotation on ~30 previews and shifts 15 Wear stickers onto a different type ramp.

## Test plan

- `:wearApp` + `:androidApp` + `:shared` compile clean.
- `PreviewThemeOverrideTest` (new, 3 tests, all pass) pins: the body theme stands down under an override; it still applies with no override; nested theming survives an override.
- Re-rendered the whole Wear catalog via `:wearApp:composePreviewRender` — **129/129 PNGs byte-identical to `main`**. No pixel movement in the default renders.
- Probed the real override path with a throwaway preview carrying `@PreviewWrapper(KotlinConfThemeCatalog::class)` — the same daemon code path (`InvokeWithOptionalWrapper`) that `?themeProvider=` takes. Three-way comparison:
  - wrapper + body theme **==** wrapper alone (the body yielded)
  - wrapper + body theme **!=** body theme alone (the override won)

  It rendered in KotlinConf purple and JetBrains Mono instead of the body's stock Wear theme. Probe deleted afterwards; the unit test is the permanent guard.

Note the switcher only becomes live on the published site once `design-artifacts/confetti-wear` re-exports.

https://claude.ai/code/session_01LhWssk26QYPNjkx7vcbKC6

---
_Generated by [Claude Code](https://claude.ai/code/session_01LhWssk26QYPNjkx7vcbKC6)_

### DIFF
--- a/androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiThemeCatalogs.kt
+++ b/androidApp/src/main/java/dev/johnoreilly/confetti/ui/ConfettiThemeCatalogs.kt
@@ -22,47 +22,67 @@ import ee.schimke.composeai.preview.ThemeCatalog
  * with, so a provider that read the ambient mode would make half the options no-ops. Each wraps a
  * production theme function — [ConfettiTheme] for the three colour schemes, [ConferenceMaterialTheme]
  * for the seeded app-root theme — so these can't drift from what the app renders.
+ *
+ * Every provider goes through [Scheme] or [ConferenceSeed], both of which mark the theme they
+ * installed as an override so the preview body's own theme stands down instead of shadowing it —
+ * see [PreviewThemeOverrideInstalled]. Without that mark the switcher renders identical pixels for
+ * every entry here, which is exactly the bug this indirection exists to prevent.
  */
+
+/** One of [ConfettiTheme]'s three colour schemes, installed as a preview theme override. */
+@Composable
+private fun Scheme(
+    dark: Boolean,
+    androidTheme: Boolean = false,
+    disableDynamicTheming: Boolean = true,
+    content: @Composable () -> Unit,
+) =
+    ConfettiTheme(
+        darkTheme = dark,
+        androidTheme = androidTheme,
+        disableDynamicTheming = disableDynamicTheming,
+    ) {
+        PreviewThemeOverrideInstalled(content)
+    }
+
 @ThemeCatalog(name = "Confetti brand Light", group = "Confetti")
 class ConfettiBrandLightThemeCatalog : PreviewWrapperProvider {
     @Composable
-    override fun Wrap(content: @Composable () -> Unit) =
-        ConfettiTheme(darkTheme = false, disableDynamicTheming = true, content = content)
+    override fun Wrap(content: @Composable () -> Unit) = Scheme(dark = false, content = content)
 }
 
 @ThemeCatalog(name = "Confetti brand Dark", group = "Confetti")
 class ConfettiBrandDarkThemeCatalog : PreviewWrapperProvider {
     @Composable
-    override fun Wrap(content: @Composable () -> Unit) =
-        ConfettiTheme(darkTheme = true, disableDynamicTheming = true, content = content)
+    override fun Wrap(content: @Composable () -> Unit) = Scheme(dark = true, content = content)
 }
 
 @ThemeCatalog(name = "Android Light", group = "Android")
 class AndroidLightThemeCatalog : PreviewWrapperProvider {
     @Composable
     override fun Wrap(content: @Composable () -> Unit) =
-        ConfettiTheme(darkTheme = false, androidTheme = true, content = content)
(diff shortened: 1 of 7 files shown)


yschimke:
Both items in the preview-diff comment above are pre-existing / benign — checked so a reviewer doesn't have to.

**`MainActivity` render failures (×2)** — pre-existing, not a regression. There is no baseline PNG for them either; both commits 404:

```
renders/wearApp/activity__MainActivity.png @ 5384c7ad (base) → 404
renders/wearApp/activity__MainActivity.png @ d4e0861a (head) → 404
```

Reproduces locally on this branch with `KoinApplication has not been started`, thrown from `DefaultWearAppComponent.getKoin` before any theming runs — an activity capture that needs Koin started, unrelated to this change.

**`HomeListViewKotlinConf` changed** — a stitching artifact in the `@ScrollingPreview(LONG)` capture, and the head render is the cleaner of the two. The base image draws a duplicated, half-clipped fourth "Friday" chip below the three real ones; the head image ends correctly at Wednesday / Thursday / Friday. Nothing else differs.

This change can't add or remove a list item: `HomeScreen.kt` is untouched (`git diff origin/main` on it is empty), and the only path it touches for this preview is `ConfettiThemeFixed`, which gained a `LocalPreviewThemeOverride` read that is `false` here and falls straight through to the same `MaterialTheme(colors, typography)` call as before.

The other 199 variants across 94 functions are unchanged, which is the expected result — this change is deliberately pixel-neutral for the default theme. It only alters rendering when a `@ThemeCatalog` / `@WearThemeCatalog` provider is wrapped around a preview, which the baked catalog render doesn't do.

---

PR #1774, opened 2026-08-03
chore(deps): update compose-ai-tools to v0.19.25

yschimke:
Bumps compose-ai-tools from `0.19.18` to `0.19.25` — the renderer behind the published `/confetti-mobile/` and `/confetti-wear/` design catalogs.

## Changes

- `gradle/libs.versions.toml`: `composeai-preview = "0.19.25"`
- `.github/workflows/design-artifacts.yml`: `install@v0.19.25`
- `.github/workflows/compose-preview.yml`: both `apply@` refs to `v0.19.25`

## Renovate grouping

This repo pins compose-ai-tools in **two** places — the `composeai-preview` catalog key and the `@v0.19.x` action refs — but `renovate.json` has no rule tying them together. They're separate updates to Renovate, so they land in separate PRs and the CLI can sit at one release while the action code driving it sits at another. compose-ai-tools ships the CLI, the Gradle plugin and the composite actions as one release, so that skew isn't harmless.

Added a `packageRules` entry grouping both under `compose-ai-tools`, matching what the sibling repos publishing to preview.coo.ee already do. Automerge is deliberately left off — that's your call on this repo, not something I wanted to switch on.

Happy to drop the renovate.json change and keep this to the version bump if you'd rather keep the config untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BEAnP7catuBrUzog96WT1C)_

### DIFF
--- a/.github/workflows/compose-preview.yml
+++ b/.github/workflows/compose-preview.yml
@@ -88,7 +88,7 @@ jobs:
           cache-read-only: ${{ github.event_name == 'pull_request' }}
       - name: Warm Gradle wrapper
         run: ./gradlew help
-      - uses: yschimke/compose-ai-tools/.github/actions/apply@v0.19.18
+      - uses: yschimke/compose-ai-tools/.github/actions/apply@v0.19.25
         with:
           # Pin the render CLI to the applied Gradle plugin version (single
           # source of truth: composeai-preview in gradle/libs.versions.toml)
@@ -131,7 +131,7 @@ jobs:
           cache-read-only: ${{ github.event_name == 'pull_request' }}
       - name: Warm Gradle wrapper
         run: ./gradlew help
-      - uses: yschimke/compose-ai-tools/.github/actions/apply@v0.19.18
+      - uses: yschimke/compose-ai-tools/.github/actions/apply@v0.19.25
         with:
           cli-version: catalog
           catalog-key: composeai-preview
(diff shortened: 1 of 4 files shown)


yschimke:
The failing **Render previews** check is not caused by this bump — it reproduces on `main` with the old version.

| ref | action version | A11y previews | Render previews |
|---|---|---|---|
| `main` @ [`c27093e`](https://github.com/joreilly/Confetti/commit/c27093ea80c13fcd7349e5e32b42adf278844623) ([run](https://github.com/joreilly/Confetti/actions/runs/30833059514)) | `apply@v0.19.18` | ✅ | ❌ |
| this PR @ `f3c35c5` ([run](https://github.com/joreilly/Confetti/actions/runs/30855768732)) | `apply@v0.19.25` | ✅ | ❌ |

Same job, same step, same pass/fail split on both versions. `c27093e` ("Layout Improvements" #1773) is the commit this branch is based on, and Compose Preview went red there — the two runs before it on `main` were green.

The signature in both runs is identical: `composePreviewDiscover` and `composePreviewRenderAll` fail across all six modules (`:androidApp`, `:compose-desktop`, `:compose-web`, `:proto`, `:shared`, `:wearApp`), `_previews.json` comes back empty, the comment step logs `nothing to compare, skipping`, and the pipeline exits 2. Because discovery fails for every module rather than a subset, this reads as a build/configuration failure rather than a preview regression — the render never gets far enough to produce or diff pixels.

So there's no render-diff evidence to show for this PR, which is unfortunate given the bump spans 0.19.21–0.19.25 where most of the changes are theme rendering. Once the underlying break is fixed the diff will be meaningful.

Happy to dig into the actual `composePreviewDiscover` failure as a separate PR if useful — it's independent of this bump and wants its own change.

---

PR #1801, opened 2026-08-13
Remove schema CacheControl directives to prevent cache read OutOfMemoryError

marcellogalhardo:
Remove @CacheControl annotations from extra.graphqls to stop Apollo from installing SchemaCoordinatesMaxAgeProvider.

When @CacheControl directives are defined on types in extra.graphqls, Apollo's normalized cache compiler plugin generates a maxAges map and attaches CacheControlCacheResolver to cache reads. During cache reads for large payloads such as GetConferenceDataQuery, SchemaCoordinatesMaxAgeProvider.getMaxAge() executes recursive type lookups on every single field. Each lookup calls allImplements(), creating repeated LinkedHashSet, ArrayList, and distinct() allocations for thousands of nodes. This excessive allocation pressure rapidly exhausts heap memory and triggers OutOfMemoryError on Android.

Confetti already configures maxStale(Duration.INFINITE) in ApolloClientCache and manages cache freshness explicitly via FetchPolicy (CacheFirst vs NetworkOnly). Removing @CacheControl clears the maxAges map, completely bypassing SchemaCoordinatesMaxAgeProvider with zero loss of caching functionality.

**Disclosure:** this PR contains code produced by generative Al.

### DIFF
--- a/shared/src/commonMain/graphql/extra.graphqls
+++ b/shared/src/commonMain/graphql/extra.graphqls
@@ -1,8 +1,7 @@
 extend schema
-@link(url: "https://specs.apollo.dev/kotlin_labs/v0.3")
 @link(
-    url: "https://specs.apollo.dev/cache/v0.3",
-    import: ["@typePolicy", "@fieldPolicy", "@cacheControl", "@cacheControlField"]
+    url: "https://specs.apollo.dev/kotlin_labs/v0.3",
+    import: ["@typePolicy", "@fieldPolicy"]
 )
 
 extend type Venue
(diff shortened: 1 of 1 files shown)


BoD:
For posterity (and, this really ought to be a comment in the code or in some README!):

We have a [custom fetch policy](https://github.com/joreilly/Confetti/blob/main/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt#L39) in Confetti, which:

- tries the cache
- if it's a hit and not stale according to the max age defined in the schema -> emit that
- otherwise -> try the network
- if the network succeeds -> emit that
- otherwise -> emit the cached value, if any, even stale - because stale is better than nothing
- if there was no cached value -> emit an error


The `maxStale(Duration.INFINITE)` that the assistant thought meant the max ages were ignored doesn't actually mean that - it's just a way to still get stale values from the cache and implement that custom logic.

----
My preference would be to continue using cache control and understand and fix (in Apollo Kotlin) any perf problems that it may provoke. Of course that can take time and I have nothing against reverting cache control while this is looked into. If we want to do that, this PR should also remove the custom fetch policy.

marcellogalhardo:
I am closing this PR for now. It seems I can no longer reproduce the issue, with or without these changes. While these changes fixed the problem for me originally, something else also resolved it; likely one of the GraphQL changes in our recent optimizations.

PR #1805, opened 2026-08-13
Remove Startup Loading Delays and Improve Cache-First Offline Operation

marcellogalhardo:
Conference attendees often encounter congested venue networks with poor connectivity. When opening the app between presentations, users need to see the schedule immediately without waiting for network responses.

Previously, `CacheFirst` treated cached queries as stale and blocked screen transitions until a network request completed on launch. This caused multi-second startup delays and loading errors when offline.

This change updates `CacheFirstInterceptor` to emit cached data from local storage immediately with zero delay, then trigger a background network sync asynchronously.

As a result, screens render instantly using cached data, while background requests update local storage silently and ignore network failures when offline.

### DIFF
--- a/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
+++ b/shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
@@ -7,18 +7,23 @@ import com.apollographql.apollo.api.Operation
 import com.apollographql.apollo.interceptor.ApolloInterceptor
 import com.apollographql.apollo.interceptor.ApolloInterceptorChain
 import com.apollographql.cache.normalized.FetchPolicy
-import com.apollographql.cache.normalized.cacheInfo
-import com.apollographql.cache.normalized.errorsAsException
 import com.apollographql.cache.normalized.fetchFromCache
 import com.apollographql.cache.normalized.fetchPolicy
 import com.apollographql.cache.normalized.fetchPolicyInterceptor
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.flow
-import kotlinx.coroutines.flow.single
+import kotlinx.coroutines.flow.singleOrNull
+import kotlinx.coroutines.launch
+
+private val backgroundScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
 
 /**
- * Use our custom fetch policy interceptor for the [FetchPolicy.CacheFirst] policy, and the Apollo built-in interceptors
- * for the other policies.
+ * Configure execution options to use [CacheFirstInterceptor] for [FetchPolicy.CacheFirst]
+ * and default Apollo interceptors for other policies.
  */
 fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when (fetchPolicy) {
     FetchPolicy.CacheFirst -> {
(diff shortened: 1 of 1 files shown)


shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
```diff
@@ -31,32 +36,39 @@ fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when
...
         request: ApolloRequest<D>,
         chain: ApolloInterceptorChain
     ): Flow<ApolloResponse<D>> = flow {
-        val cacheResponse = chain.proceed(request.newBuilder().fetchFromCache(true).build()).single()
-        if (cacheResponse.cacheInfo!!.isCacheHit && !cacheResponse.cacheInfo!!.isStale) {
+        val cacheResponse = try {
+            chain.proceed(request.newBuilder().fetchFromCache(true).build()).single()
+        } catch (e: Exception) {
+            null
+        }
```
BoD: I'm not expecting exceptions being thrown here - did you add the `try` because you witnessed one?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
```diff
@@ -31,32 +36,39 @@ fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when
...
+
+        if (cacheResponse?.data != null) {
+            // Emit cached data immediately to prevent startup loading delay
             emit(cacheResponse)
```
BoD: This is a matter of preference but in the online case, this has the potential to show stale data for a fraction of a second, and then fresh, different data - a content flicker, where you see something and before you can click on it, it's gone, which can be annoying.

Tradeoff: content appears immediately vs content may flickers

Maybe we could do this only if we know we're offline (`ConnectivityManager`)?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
```diff
@@ -31,32 +36,39 @@ fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when
...
+
+            // Refresh cache in background. Ignore network errors to keep cached data visible offline
+            backgroundScope.launch {
+                try {
+                    chain.proceed(request).collect()
+                } catch (e: Exception) {
+                    // Suppress network errors to keep valid cached data visible
+                }
```
BoD: Also not expecting exceptions here.

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
```diff
@@ -31,32 +36,39 @@ fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when
...
 /**
- * Returns by priority:
- * - the cached data if it's fresh
- * - the network data if it could be fetched
- * - the stale cached data
+ * Interceptor for [FetchPolicy.CacheFirst] operations.
+ *
+ * Emits cached data immediately to show the UI without delay.
+ * Updates the cache in the background. Network errors during background
+ * sync are ignored to keep valid cached data visible.
```
BoD, line 43: As discussed on Slack I'd love to keep staleness into consideration for what's shown in the UI (which is currently, basically, only allow showing stale data if the network fails, with "stale data" meaning age > the max ages defined in the schema)

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
marcellogalhardo: Fair tradeoff. My reasoning for leaning toward immediate cache rendering here:

- Schedule updates do happen, but infrequent enough during an event that user tapping a moving item in that sub-second window seems very small (open to data if we have metrics!).
- When updates happen, Compose stable keys (`key = { it.id }`) minimize visible flicker by diffing items in place rather than doing a full redraw.
- Often conference Wi-Fi shows full signal, but actual throughput is dead. A connectivity check would report online, leaving user blocked on a loading spinner while moving between rooms.

I experienced this firsthand at KotlinConf: the phone reported connected Wi-Fi, but with venue network congestion, Confetti stayed stuck on a loading spinner when I needed the schedule. I had to use the official KotlinConf app.

Because of this, my intuition is that immediate cache rendering offers a better UX during a conference.

What do you think?

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
```diff
@@ -31,32 +37,29 @@ fun <T> MutableExecutionOptions<T>.fetchPolicy(fetchPolicy: FetchPolicy) = when
...
-        val cacheResponse = chain.proceed(request.newBuilder().fetchFromCache(true).build()).single()
-        if (cacheResponse.cacheInfo!!.isCacheHit && !cacheResponse.cacheInfo!!.isStale) {
+        val cacheResponse = chain.proceed(request.newBuilder().fetchFromCache(true).build()).singleOrNull()
+
+        if (cacheResponse?.data != null) {
+            // Emit cached data immediately to prevent startup loading delay.
             emit(cacheResponse)
+
+            // Refresh cache in background.
+            backgroundScope.launch { chain.proceed(request).collect() }
```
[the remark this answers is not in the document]
marcellogalhardo: cc: @BoD

shared/src/commonMain/kotlin/dev/johnoreilly/confetti/fetchPolicy.kt
marcellogalhardo: I replaced `single()` by `singleOrNull()` to avoid an `Exception`.

BoD:
So sorry, it looks like I clicked 'Merge' by mistake. Revert [here](https://github.com/joreilly/Confetti/pull/1831).

PR #1812, opened 2026-08-14
Fix sign-out and navigation loss on configuration changes

marcellogalhardo:
DefaultAppComponent initialized user state to null. When Android restored the navigation stack during a configuration change, child components received a null user before the authentication StateFlow emitted. Because the restored stack configuration already contained the user ID, subsequent user updates did not trigger component recreation, leaving the UI signed out and bookmark mutations failing authorization.

Additionally, DefaultAppComponent.init executed conference navigation unconditionally, destroying the restored child backstack.

Initialize user state synchronously from the authentication StateFlow on component creation and restrict initial navigation to Loading state or deep links.


joreilly:
nice find

PR #1820, opened 2026-08-16
Add embedded field policies for nested sub-objects in extra.graphqls

marcellogalhardo:
Configure embeddedFields for types without independent primary keys (Session.links, Speaker.socials, PartnerGroup.partners, and connection pageInfo).

Without embedded field policies, Apollo normalized cache assigns composite generated keys and normalizes sub-objects into standalone cache records. Embedding them directly within their parent record reduces cache record count, memory footprint, and database lookups.

**Disclosure:** this PR contains code produced by generative Al.

### DIFF
--- a/shared/src/commonMain/graphql/extra.graphqls
+++ b/shared/src/commonMain/graphql/extra.graphqls
@@ -9,7 +9,7 @@ extend type Venue
 @typePolicy(keyFields: "id")
 
 extend type Session
-@typePolicy(keyFields: "id")
+@typePolicy(keyFields: "id", embeddedFields: "links")
 @cacheControl(maxAge: 14400)
 
 extend type Bookmarks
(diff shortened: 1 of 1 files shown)


shared/src/commonMain/graphql/extra.graphqls
```diff
@@ -9,17 +9,26 @@ extend type Venue
...
 @cacheControl(maxAge: 14400)
 
+extend type PartnerGroup
+@typePolicy(embeddedFields: "partners")
+
+extend type SessionConnection
+@typePolicy(embeddedFields: "pageInfo")
+
+extend type SpeakerConnection
+@typePolicy(embeddedFields: "pageInfo")
```
BoD, line 30: These 2 are interesting: if we used the cache's pagination feature they would be embedded automatically. At the moment we don't use it in Confetti.

BoD (approved):
LGTM!

BoD:
Thinking out loud: I wonder if there were some hints we could add to the Apollo IDE plugin to suggest to embed certain fields like those. Not sure what rule it would use - is it 'fields whose type doesn't have an id/cache key'? Food for thought cc @martinbonnin

PR #1840, opened 2026-08-17
Modernise Renovate config and make release trains atomic

yschimke:
Triggered by okhttp 5.5.0 landing as three separate Renovate PRs (`okhttp`, `okhttp-coroutines`, `logging-interceptor`). All three sit on the same classpath in `backend/service-import`, so that classpath is skewed until the last of the three merges. Each was declared with its own hardcoded version, so Renovate had no reason to group them.

## Catalog

Renovate updates catalog entries that share a version ref in a single branch — that is why `roborazzi` already lands as one PR. So the durable fix for the split trains is at the catalog level, not in `renovate.json`:

| train | was | now |
|---|---|---|
| okhttp ×3 | hardcoded 5.4.0 | `okhttp = "5.5.0"` |
| ktor ×5 | hardcoded 3.5.2 | `ktor` ref |
| scrimage ×2 | hardcoded 4.6.7 (split as #1737 / #1738) | `scrimage` ref |
| androidx.lifecycle | 3 refs at the same version | one `androidx-lifecycle` ref |
| coroutines-play-services | own ref | `kotlinx-coroutines` |
| protolayout-material3 | hardcoded 1.4.1 | `protolayout` ref |
| spring-boot-gradle-plugin | hardcoded 3.5.16 | `spring` ref |
| stately-common | hardcoded 2.1.0 | `stately` ref |

**okhttp is the only version that changes** (5.4.0 → 5.5.0). Every other value is identical to what was already resolving, so nothing else moves.

The three open okhttp Renovate PRs can be closed.

## renovate.json

- `config:base` → `config:recommended`. `config:base` has been a deprecated alias for years; Renovate migrates it silently, but the file was documenting an obsolete preset.
- `:semanticCommitsDisabled`. Commit titles were inconsistent — `chore(deps): update gradle to v9.6.1` in #1711–#1717, plain `Update dependency …` in #1745+. That is Renovate's autodetection flipping as the `Bump Android version to …` release commits push semantic ones out of the sample window. Pinning it to the dominant plain style keeps titles stable.
- `configMigration: true`, so deprecated config is flagged by a PR instead of rotting the way `config:base` did.
- Group every `com.squareup.okhttp3` artifact. The shared version ref already covers the catalog entries; this keeps the guarantee for any okhttp3 artifact declared outside it.
- Group non-major GitHub Actions updates. One round of action majors landed as six PRs (#1762–#1766). Majors still get their own PR so the breaking-change notes get read. Ordered **before** the compose-ai-tools rule so that rule still wins for its action half.
- Hold `kotlinx-datetime` on the `-0.6.x-compat` line. Plain `0.8.0` sorts *above* `0.8.0-0.6.x-compat` in Maven ordering, so Renovate would eventually offer what looks like a routine bump and silently drop the compat variant that `apollo-adapters-kotlinx-datetime` needs. Delete the rule once the datetime consumers move to the new `Instant` API.
- Refreshed the stale `@v0.19.x` reference in the compose-ai-tools rule description — the workflows are on `@v1.9.0`.

## Verification

- `renovate-config-validator` passes against latest Renovate (CI runs the same check in `.github/workflows/verify.yml`).
- `:backend:service-import:compileKotlinJvm` and `:backend:service-graphql:compileKotlin` build on okhttp 5.5.0.

## Noted, deliberately not changed

- **Spring skew:** `spring-web = 6.2.19` but `spring-webflux = 7.0.8` — Framework 7 needs Boot 4, and this repo is on Boot 3.5.16. Both entries are unused (only `libs.spring.boot.starter.webflux` is referenced), as are `spring-boot`, `spring-boot-autoconfigure` and `spring-boot-starter-logging`. Deleting them would remove the skew and the update noise, but that is a separate call.
- `"schedule": ["at any time"]` on the compose-ai-tools rule is a no-op: nothing sets a repo-level `schedule` or `minimumReleaseAge`, so there is no batching window to escape. Left as intent documentation.
- `"agp"` in the AGP rule's `matchPackageNames` matches nothing — Renovate keys on `group:artifact`, not the catalog version key. The other two patterns do the work; left alone rather than risk unblocking AGP.
- `config:best-practices` would add GitHub Actions digest pinning (a real supply-chain win), but it rewrites all ~16 action refs to SHAs, so it is worth deciding on separately.
- `gradle/gradle-build-action@v3.5.0` is deprecated in favour of `gradle/actions/setup-gradle`, which the other 10 call sites already use.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,6 +4,7 @@ ksp = "2.3.10"
 kotlinx-coroutines = "1.11.0"
 kotlinx-datetime = "0.8.0-0.6.x-compat"
 kotlinx-serialization = "1.11.0"
+ktor = "3.5.2"
 apollo-kotlin-execution = "0.1.2"
 tapmoc = "0.4.2"
 
@@ -32,12 +33,10 @@ kermit = "2.1.0"
 kmmbridge = "1.2.1"
 kotlin-csv = "1.11.0"
 koin = "4.2.2"
-kotlinx-coroutines-play-services = "1.11.0"
-lifecycle = "2.11.0"
-lifecycle-livedata-ktx = "2.11.0"
 materialkolor = "5.0.0"
 multiplatform-settings = "1.3.0"
 nav-compose = "2.9.8"
+okhttp = "5.5.0"
 okio = "3.18.1"
 permissions = "0.20.1"
 haze = "1.7.2"
@@ -46,6 +45,7 @@ stately = "2.1.0"
 protolayout = "1.4.1"
 robolectric = "4.16.1"
 room = "2.8.4"
+scrimage = "4.6.7"
 tiles-tooling-preview = "1.6.1"
 wear = "1.4.0"
 wear-watchface = "1.3.0"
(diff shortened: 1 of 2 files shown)


yschimke:
No fix pushed — these two render failures are pre-existing and unrelated to this PR. The identical `MainActivity` (androidApp) / `MainActivity` (wearApp) failures appear on every recent preview run, including #1836 and #1831, and this PR renders 214 variants unchanged with zero visual diffs (expected, since the only version that moves is okhttp 5.4.0 → 5.5.0, which is backend-only).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1867, opened 2026-08-21
Add HTTP `QUERY` support

martinbonnin:


backend/service-graphql/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ dependencies {
   implementation(platform(libs.spring.boot.bom))
 
   implementation(libs.spring.boot.starter.webflux)
+  implementation("org.springframework:spring-webflux:7.1.0-SNAPSHOT")
```
martinbonnin, line 23: I'm not a fan of adding SNAPSHOTs but I'd like to test using Google Cloud CDN and there are no Spring releases with `QUERY` yet. Feel free to revert if you're reading this and something went wrong.

martinbonnin: Welp, if you're curious, that didn't work out at all... Google cloud doesn't seem ready for it and returns 502, whether going through the CDN or not.

